PR #4, opened 2017-03-12
Testrpc

makoto:
Hi.

Your README instruction suggested that testrpc is installed locally but package.json did not have it.

https://github.com/makoto/protocol#testing


retotrinkler:
thanks

PR #5, opened 2017-03-12
Require contracts

makoto:
Hi. 

When I run the test according to your README instruction, I encountered the following 3 errors.

https://gist.github.com/makoto/8c117d741d9960a7116dcd2e4513f626

These changes should fix 2 of 3 errors. I don't think I can fix the last error as I cannot guess the correct behavior from the test and I also saw that other tests in the same test files are commented out.

Thanks.

### DIFF
--- a/test/nav.js
+++ b/test/nav.js
@@ -4,6 +4,15 @@ const BigNumber = require('bignumber.js');
 const constants = require('../utils/constants.js');
 const functions = require('../utils/functions.js');
 
+const EtherToken = artifacts.require("./EtherToken.sol");
+const BitcoinToken = artifacts.require("./BitcoinToken.sol");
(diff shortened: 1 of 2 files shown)


retotrinkler:
yeah nav and exchange need fixing..

makoto:
BTW what is the responsibility of `nav.js` ? Is it more like integration test (shorten for `naviation`)?

retotrinkler:
NAV stands for net asset value. Basically how much value all assets have together (denominated in Ether)

retotrinkler:
ah idea of nav.js is to test investing in portfolio then trading of manager then input of new prices then recalculating nav and thus performance and finally redeeming - should simulate an entire investment - trading - redeeming circle

PR #21, opened 2017-05-22
Adjust contract instantiation to match Core constructor

nud3l:
The tests in makeOrder.js and subscribe.js did not reflect the new Core.sol constructor. Adjusted the creation of contracts in the test files as in nav.js and version.js.


retotrinkler:
thank you

PR #25, opened 2017-05-23
Riskmgmtv1

### LINKED ISSUE #17: Initial Version of Risk Mgmt module
Possibility of jumping orders; There is no enforcement of manager to take best offered order

- Initial solution:
Assert that: 
`order.owner === address(liquidityProvider)`
for takeOrder

retotrinkler:
Closes #17

### DIFF
--- a/contracts/Core.sol
+++ b/contracts/Core.sol
@@ -120,7 +120,7 @@ contract Core is DBC, Owned, Shares, SafeMath, CoreProtocol {
         for (uint i = 0; i < numAssignedAssets; ++i) {
             PriceFeedProtocol Price = PriceFeedProtocol(address(module.universe.priceFeedAt(i)));
             address quoteAsset = Price.getQuoteAsset();
-            assert(referenceAsset == quoteAsset);
+            require(referenceAsset == quoteAsset);
         }
         module.subscribe = SubscribeProtocol(ofSubscribe);
         module.redeem = RedeemProtocol(ofRedeem);
@@ -144,7 +144,6 @@ contract Core is DBC, Owned, Shares, SafeMath, CoreProtocol {
     function createSharesOnBehalf(address recipient, uint shareAmount)
         pre_cond(notZero(shareAmount))
     {
-        calcSharePrice();
         allocateSlice(recipient, shareAmount);
         SharesCreated(recipient, now, shareAmount);
     }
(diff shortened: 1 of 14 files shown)


contracts/Core.sol
```diff
@@ -48,27 +51,22 @@ contract Core is DBC, Owned, Shares, SafeMath, CoreProtocol {
...
-    uint public sharePrice;
-    uint public unclaimedFees; // Combined total of Management and Performance fees earned by portfolio manager
     // Fields that are only changed in constructor
     address referenceAsset;
     address melonAsset;
     // Fields that can be changed by functions
     Modules module;
     Properties properties;
-    Calculations calculated;
+    Calculations public atLastPayout;
```
travs: Good variable name, more clear to me now :+1:

contracts/Core.sol
```diff
@@ -259,18 +269,17 @@ contract Core is DBC, Owned, Shares, SafeMath, CoreProtocol {
 
     /// Pre: Universe has been defined
     /// Post: Whether buying and selling of tokens are allowed at given exchange
-    function isWithinKnownUniverse(address onExchange, address sell_which_token, address buy_which_token)
+    function requireIsWithinKnownUniverse(address onExchange, address sell_which_token, address buy_which_token)
```
travs: Not a big deal since it works as is, but should this be a modifier I wonder?

test/makeOrder.js
```diff
@@ -161,11 +161,11 @@ contract('Net Asset Value', (accounts) => {
       });
     });
 
-    it('Create an initial order', (done) => {
+    it.skip('Create an initial order', (done) => {
       const wantedShares = new BigNumber(2e+17);
       const wantedValue = new BigNumber(2e+17);
       const expectedValue = new BigNumber(2e+17);
-
+
```
travs: oops

contracts/Core.sol
retotrinkler: :)

contracts/Core.sol
retotrinkler: ideally yes! however there is [this one time](https://github.com/melonproject/protocol/blob/c1c3a8d670fbcf2402528eb90e96c9d8e44ac0c2/contracts/Core.sol#L244-L249) where it needs to get (call) the parameteres for it first

test/makeOrder.js
retotrinkler: ah np at all - guess it happened automatically w my tool

PR #26, opened 2017-05-26
Test Share creation/annihilation, Closes #15

### LINKED ISSUE #15: Subscribe module testcase for creating and annihilation shares

travs:


contracts/Core.sol
```diff
@@ -120,11 +120,15 @@ contract Core is DBC, Owned, Shares, SafeMath, CoreProtocol {
...
-        returns (uint sharePrice)
+        returns (uint valuePerShare)
     {
-        sharePrice = 10 ** decimals * value / totalSupply;
+        valuePerShare = 10 ** decimals * value / totalSupply;
     }
 
+    /// Pre: None
+    /// Post: Share price denominated in referenceAsset in baseunit of [10 ** decimals] per Share
+    function getSharePrice() constant returns (uint price){price = atLastPayout.sharePrice;}
```
travs: @retotrinkler 
Added this to get price when there are zero shares.
May be better to modify `calcValuePerShare` so we don't have two functions.
Let me know what you think.

contracts/Core.sol
```diff
@@ -226,17 +230,18 @@ contract Core is DBC, Owned, Shares, SafeMath, CoreProtocol {
...
     {
         if (totalSupply == 0) { // Iff all coreHoldings are zero
             /* By definition for zero totalSupply of shares:
              *  sharePrice == initialSharePrice (1)
-             *  hence for actualValue == sharePrice * shareAmount / initialSharePrice == shareAmount unsing (1) above
+             *  hence for actualValue == sharePrice * shareAmount / initialSharePrice == shareAmount using (1) above
              */
-            assert(AssetProtocol(referenceAsset).transferFrom(msg.sender, this, shareAmount)); // Send msg.sender to this core
+             uint totalCost = shareAmount * getSharePrice();
+             assert(AssetProtocol(referenceAsset).transferFrom(msg.sender, this, totalCost)); // Send from msg.sender to core
```
travs: attn @retotrinkler 
Messaging you about this

contracts/Core.sol
retotrinkler: ah right it throws for empty zero shares..
atLastPayout.sharePrice can however be quite delayed
think it works for now but we should keep it in mind for later

contracts/Core.sol
retotrinkler: hmm yeah think this works

contracts/Core.sol
```diff
@@ -153,7 +150,7 @@ contract Core is DBC, Owned, Shares, SafeMath, CoreProtocol {
         uint gav = calcGav(); // Reflects value indepentent of fees
         var (managementFee, performanceFee, unclaimedFees) = calcUnclaimedFees(gav);
         uint nav = calcNav(gav, unclaimedFees);
-        uint sharePrice = calcValuePerShare(nav);
+        uint sharePrice = notZero(totalSupply) ? calcValuePerShare(nav) : initialSharePrice; // Pot division through zero requires defined amount
```
retotrinkler: handled case for empty supply

contracts/Core.sol
retotrinkler: has `real time` data

contracts/participation/Subscribe.sol
```diff
@@ -42,7 +42,8 @@ contract Subscribe is SubscribeProtocol, DBC, SafeMath, Owned {
         pre_cond(isPastZero(wantedShares))
     {
         CoreProtocol Core = CoreProtocol(ofCore);
-        uint actualValue = Core.getSharePrice();
+        var (, , , , , sharePrice) = Core.performCalculations();
+        uint actualValue = sharePrice * wantedShares;
```
retotrinkler: added multiplication

PR #29, opened 2017-06-08
Improve documentation and variable naming

retotrinkler:


contracts/Vault.sol
```diff
@@ -220,47 +221,47 @@ contract Core is DBC, Owned, Shares, SafeMath, CoreProtocol {
...
-        if (totalSupply == 0) { // Iff all coreHoldings are zero
-            /* By definition for zero totalSupply of shares:
-             *  sharePrice == initialSharePrice (1)
-             *  hence for actualValue == sharePrice * shareAmount / initialSharePrice == shareAmount using (1) above
+        if (totalSupply == 0) { // Iff all vaultHoldings are zero
+            /* By definition for zero totalSupply of shares the initial share price is defined as:
+             *  sharePrice == baseUnitsPerShare (1)
+             *  hence for totalCost == shareAmount * sharePrice / baseUnitsPerShare == shareAmount using (1) above
              */
              var (, , , , , sharePrice) = performCalculations();
```
travs: Nevermind, looks like it updates some globals so we can leave it

contracts/Vault.sol
```diff
@@ -220,47 +221,47 @@ contract Core is DBC, Owned, Shares, SafeMath, CoreProtocol {
...
     function annihilateSharesOnBehalf(address recipient, uint shareAmount)
         pre_cond(balancesOfHolderAtLeast(recipient, shareAmount))
     {
         separateSlice(recipient, shareAmount);
         SharesAnnihilated(recipient, now, shareAmount);
     }
 
-    /// Pre: Allocation: Pre-approve spending for all non empty coreHoldings of Assets
-    /// Post: Transfer ownership percentage of all assets to/from Core
+    /// Pre: Allocation: Pre-approve spending for all non empty vaultHoldings of Assets, shareAmount denominated in [base units ]
```
retotrinkler: :) my bad

contracts/Vault.sol
retotrinkler: not necessary anymore!
think all changes should be non-persistent (only constant function calls)
fixed in new pr

PR #31, opened 2017-06-08
rem unnec statement

retotrinkler:


travs:
Haha, I wasn't sure if that was necessary or not. Thanks for clearing it up! :smile:

PR #39, opened 2017-06-12
Subscribe redeem branch - Fixes #30

### LINKED ISSUE #30: Subscribe and redeem through subscribe and redeem modules
Subscribe with only Ether and redeem by receiving only Ether

travs:
- Adds ability for subscribe/redeem modules to create/annihilate shares on behalf of an investor who only uses the fund's reference asset.
- Adds tests for that functionality, with two assets in the fund
- Fixes #30


retotrinkler:
v nice!

PR #58, opened 2017-07-06
Participation

retotrinkler:


migrations/2_deploy_contracts.js
```diff
@@ -28,7 +28,7 @@ module.exports = (deployer, network, accounts) => {
     );
   }
 
-  if (network == "kovan") {
+  if (network === 'kovan') {
     deployer.deploy([
       Exchange,
       // Redeem,
```
travs: There was a conflict in this file, where I kept the [most recent change](https://github.com/melonproject/protocol/blob/master/migrations/2_deploy_contracts.js) instead of this one.

I think that is what you want, but let me know if it is the opposite 😃

travs:
Merged master into here to fix merge conflicts, and now the build breaks because master is also breaking.
I'm pretty sure the reason is the addition of other asset contracts. 
I recommend we *don't* merge with a breaking build on a feature branch, so I will create another branch from here (a35bca7) called `fix-tests`, where I will begin fixing the tests until we have a good build.

migrations/2_deploy_contracts.js
retotrinkler: yup - most change looks good esp [this](https://github.com/melonproject/protocol/blob/master/migrations/2_deploy_contracts.js#L53-L58)

PR #61, opened 2017-07-10
New tokens

travs:


retotrinkler:
oh yeah - was bit eager to merge :)

retotrinkler:
this might be a more elegant solution: -
- https://github.com/0xProject/contracts/blob/master/contracts/TokenRegistry.sol
- with [settings file](https://github.com/0xProject/contracts/blob/a6542d978f0ddb7b694bbbe777d5c0bc1bbb147e/migrations/config/token_info.ts) instead of custom contracts
- looks super useful for [deployment](https://github.com/0xProject/contracts/blob/master/migrations/3_register_tokens.ts)

would also help solve #6

retotrinkler:
created issue #63

PR #100, opened 2017-08-22
Add vault tests

travs:
@retotrinkler 
in particular, I'm wondering if [this change](https://github.com/melonproject/protocol/compare/subscribe-rebase...add-vault-tests#diff-c29a373d5ba35ade4b1045a6318cfbe9L160) makes sense.


retotrinkler:
good

PR #105, opened 2017-08-25
Subscribe rebase into develop

travs:


contracts/governance/Version.sol
```diff
@@ -24,40 +24,64 @@ contract Version is DBC, Owned {
...
+    // PRE, POST, INVARIANT CONDITIONS
+
+    function isInHistory(uint id) constant returns (bool) { return 0 <= id && id < nextVaultId; }
+
     // CONSTANT METHODS
 
-    function getMelonAsset() constant returns (address) { return MELON_ASSET; }
-    function getLastVaultId() constant returns (uint) { return lastVaultId; }
     function getVault(uint id) constant returns (address) { return vaults[id]; }
+    function hasVault(address mgr) constant returns (bool) {
```
retotrinkler: would call `mgr`; `ofManager` as in address of manager

contracts/governance/Version.sol
```diff
@@ -100,4 +120,14 @@ contract Version is DBC, Owned {
         delete vaults[id];
         VaultUpdated(id);
     }
+
+   	function getVaults(uint start)
+        constant
+        returns (address[1024] allVaults)
+    {
+        for(uint ii = 0; ii < 1024; ii++){
```
retotrinkler: would do single `i` for enumeration

PR #130, opened 2017-09-12
New event params

travs:
Needed for integration into front.
Can quickly make a new release as soon as this is merged.


contracts/datafeeds/DataFeed.sol
```diff
@@ -35,7 +35,7 @@ contract DataFeed is DataFeedInterface, AssetRegistrar {
     // PRE, POST, INVARIANT CONDITIONS
 
     function isDataSet(address ofAsset) constant returns (bool) { return dataHistory[getLastUpdateId()][ofAsset].timestamp > 0; }
-    function isDataValid(address ofAsset) internal returns (bool) { return now - dataHistory[getLastUpdateId()][ofAsset].timestamp <= VALIDITY; }
+    function isDataValid(address ofAsset) constant returns (bool) { return now - dataHistory[getLastUpdateId()][ofAsset].timestamp <= VALIDITY; }
```
retotrinkler: `PRE, POST, INVARIANT CONDITIONS`
by convention should be `internal`
if they are not internal they should be listed under `CONSTANT FUNCTIONS`

contracts/datafeeds/DataFeedInterface.sol
```diff
@@ -10,6 +10,8 @@ contract DataFeedInterface {
 
     // Get registartion specific information
     function isDataSet(address ofAsset) constant returns (bool) {}
+    function isDataValid(address ofAsset) constant returns (bool) {}
+    function isValid(address ofAsset) constant returns (bool) {}
```
retotrinkler: @travs `isValid` should be listed under comment `// Get asset specific information` not `// Get registartion specific information`

retotrinkler: @travs `isDataValid` and `isValid` is the same function - different only in different pre_conditions!
that's why above function was internal in the first place.
so essentially all the changes in DataFeed and DataFeedInterface should be reverted

contracts/datafeeds/DataFeed.sol
travs: ok, changed back and also changed `isDataSet` to internal

contracts/Vault.sol
```diff
@@ -472,7 +472,7 @@ contract Vault is DBC, Owned, Shares, VaultInterface {
             lastFeedUpdateTime: module.pricefeed.getLastUpdateTimestamp(),
             timestamp: now
         });
-        SubscribeRequest(msg.sender, now, numShares);
+        SubscribeRequest(msg.sender, thisId, now, numShares);
         nextRequestId++;
         return thisId;
```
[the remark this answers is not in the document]
travs: got it, should be changed in latest commit

contracts/datafeeds/DataFeedInterface.sol
travs: done, and noticed that `// Get asset specific information` was specified twice for no obvious reason, so merged under one comment. I can change that back if there was some reason I don't understand

PR #132, opened 2017-09-12
Fix visibility issues

travs:


contracts/datafeeds/DataFeedInterface.sol
```diff
@@ -20,18 +17,18 @@ contract DataFeedInterface {
...
     function getInterval() constant returns (uint) {}
     function getValidity() constant returns (uint) {}
     function getLastUpdateId() constant returns (uint) {}
     function getLastUpdateTimestamp() constant returns (uint) {}
-    // Get asset specific information
-    function getPrice(address ofAsset) constant returns (uint) {}
-    function getInvertedPrice(address ofAsset) constant returns (uint) {}
-    function getReferencePrice(address ofBase, address ofQuote) constant returns (uint) {}
-    function getTimestamp(address ofAsset) constant returns (uint) {}
-    function getData(address ofAsset) constant returns (uint, uint) {}
```
retotrinkler: the logic of this Interface is:
- Contract A 
   - Contract A **general information**; What differentiates this contract from other contracts w similar bytecode; Eg values passed in as constructor arguments or usage indicators
   - Contract A **specific information**; Eg information about Asset
- Contract B 
   - Contract B **general information**; What differentiates this contract from other contracts w similar bytecode; Eg values passed in as constructor arguments or usage indicators
   - Contract B **specific information**; Eg information about Price

contracts/Vault.sol
```diff
@@ -451,17 +451,17 @@ contract Vault is DBC, Owned, Shares, VaultInterface {
...
         pre_cond(isPastZero(incentiveValue))
-        pre_cond(module.pricefeed.isDataValid(MELON_ASSET))
+        pre_cond(module.pricefeed.isValid(MELON_ASSET))
         pre_cond(module.participation.isSubscribeRequestPermitted(
             msg.sender,
             numShares,
             offeredValue
         ))
-        returns(uint256)
+        returns (uint256 id)
```
retotrinkler: even cleaner would be to substitute `id` w `idCounterOfRequests`
with `idCounterOfRequests++` at beginning of function body

travs: then we miss entry `0` don't we?

contracts/datafeeds/DataFeedInterface.sol
travs: ok I can add comments to specify which contracts they fall under; sound reasonable?

travs: and move those events back to where they were of couse

contracts/Vault.sol
retotrinkler: these kinds of counters starting from one was always the case, and used for all similar datatypes throught melon protocol
- they are in line w https://github.com/melonproject/protocol/blob/develop/contracts/Vault.sol#L189
- benefit of: if idCounterOfRequests == 0 then its clear that there haven't been any requests, there is no entry hidden behind id == 0

retotrinkler: better name for `idCounterOfRequests` would thus be `numberOfRequests`

travs: Ok sure.
I remember discussing this and agreeing to make the change to the current, zero-indexed form ([issue](https://github.com/melonproject/protocol/issues/103)), but I'm happy to change them all back if you wish.

I think the one you reference from the vault starts from zero as well, so I would change that one back to one-indexed as well of course.
Opened [this issue](https://github.com/melonproject/protocol/issues/133) and self-assigned.

retotrinkler: one is `id` the other is `number` of requests
- its fine for `id`s to start from zero
- `number` should start from one!

PR #136, opened 2017-09-12
Change takeOrder -> takeOffer for consistency

jennazenk:
Changed takeOrder to takeOffer to be consistent w the rest of the new wording that uses offer instead of orders.


jennazenk:
Closing this one since we agree on the order label

PR #154, opened 2017-09-26
Minimal working deployment with dapp

travs:
Still many parameters to adjust/abstract, and have differing deployments depending on network config, but this is a start. 

Can do `npm run deploy` and it should work for deployment.

Also added submodules `ds-test` and `ds-token`, but these are not yet used in the protocol contracts.

Will be adding tests next, since the test command now is just a skip.


deploy-all.js
```diff
@@ -0,0 +1,158 @@
...
+        token.url,
+        mockBytes,
+        mockChainId,
+        mockAddress,
+        mockAddress,
+      ).send(opts);
+    });
+
+    // write out to JSON
+    const addressBook = {
```
travs: need to change this to add an object for each network we deploy to

PR #159, opened 2017-09-27
Dapp first tests

### LINKED ISSUE #112: npm run deploy (or similar) to create a proper version
Every deployment/release must:
- Have an unique version number. Counting up patch number seems fine to me.
- Have this version number as a git-tag (again: having a lot of git tags is not too bad)
- Have this same version as npm version

Here is some inspiration from melon.js: https://github.com/melonproject/melon.js/blob/master/package.json#L29 (also check "version" and "postversion" commands)

--> It should be super trivial to find the corresponding protocol source to a version on npm.
Now it's almost impossible :(

retotrinkler:
### Added
- Deployement specific network environments, inputted as a parameter

### Fixed
- #112


deploy-all.js
```diff
@@ -122,23 +122,37 @@ async function main() {
...
+    for(const assetSymbol of assetsToRegister) {
+      console.log(`Registering ${assetSymbol}`);
+      const thisToken = tokenInfo[networkName].filter(token => token.symbol === assetSymbol)[0];
       await datafeed.methods.register(
-        token.address,
-        token.name,
-        token.symbol,
-        token.decimals,
-        token.url,
+        thisToken.address,
```
retotrinkler: liked `token` better then `thisToken` - as [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this) as a keyword in js

deploy-all.js
```diff
@@ -11,12 +11,12 @@ function getPlaceholderFromPath(libPath) {
...
 }
 
-async function main() {
+async function deployKovan() {
   try {
     const networkName = 'kovan';
     const web3 = new Web3(new Web3.providers.HttpProvider('http://localhost:8545'));
     const accounts = await web3.eth.getAccounts();
-    const opts = { from: accounts[0], gas: 6900000 };
+    const opts = { from: accounts[0], gas: 6900000, gasPrice: 100000000000 };
```
retotrinkler: for those literals would do config file - eg something like [here](https://github.com/melonproject/microservices/blob/master/src/conf.js)

w configs dependent on `networkArgument`

PR #171, opened 2017-10-02
Rename getFund(ofManager) to getFundForManager

jennazenk:


retotrinkler:
@jennazenk if that works for u and tests pass - then it can be merged

PR #181, opened 2017-10-08
Faster initial supscription

retotrinkler:
### Changed
- Allow first supscription w/o waiting time; 
- Prevent redeem w not sufficient shares

### Removed
- Dapp files


retotrinkler:
fixed in 0.3.8

PR #202, opened 2017-10-20
Add multisig

GauthamN:


retotrinkler:
should be merged into _develop_

PR #278, opened 2017-12-04
Fix 238

travs:
### Changed

- use ds-math instead of safeMath.sol

**Note**: one of the `thirdparty/` contracts still uses `safeMath` (`EtherDelta.sol`) so we still need it in the repo if we plan on using that one


travs:
Closed in favour of #288

PR #317, opened 2017-12-12
Ava

GauthamN:

Copy/paste from CHANGELOG.md is good

### Added
- Config for ava tests
- Unit tests for riskmgmt, compliance, version

### Changed
- fundShares tests to ava.js
- RMMakeOrders to allow the equalto case

### Removed
- Jasmine dependancy


schmidsi:
Just curious: What convinced you to use ava?

GauthamN:
Think the concurrent running of tests. Although doesn't offer much difference in integration tests, think helps speeding up unit tests

schmidsi:
Yes, parallelisation is a big plus for Ava. Do you know if the expectation syntax is compatible with jasmine/jest? Would be great to share and reuse some expectation across projects. It's also helpful to use similar tools along the way.

GauthamN:
Yes, custom assertions can be used with ava. Also group testing: https://github.com/sheerun/ava-spec

PR #376, opened 2018-02-06
Before live

GauthamN:

### Added.
- RestrictedShares contract that restricts secondary trading of shares
- Test for Fund with restricted shares

### Fixed
- Bug  in quantityHeldInCustodyOfExchange

### DIFF
--- a/src/Fund.sol
+++ b/src/Fund.sol
@@ -251,14 +251,13 @@ contract Fund is DSMath, DBC, Owned, RestrictedShares, FundInterface, ERC223Rece
                 module.pricefeed.getLastUpdateId() >= add(requests[id].atUpdateId, 2)
             )
         )   // PriceFeed Module: Wait at least one interval time and two updates before continuing (unless it is the first investment)
-         // PriceFeed Module: No recent updates for fund asset list
-    {
-        // sharePrice quoted in QUOTE_ASSET and multiplied by 10 ** fundDecimals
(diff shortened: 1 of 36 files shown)


src/Fund.sol
```diff
@@ -731,14 +732,15 @@ contract Fund is DSMath, DBC, Owned, Shares, FundInterface, ERC223ReceivingContr
             if (sellQuantity == 0) {
                 exchangeIdsToOpenMakeOrderIds[i][ofAsset] = 0;
             }
-            if (!exchanges[i].isApproveOnly) {
-                totalSellQuantity += sellQuantity;
+            totalSellQuantity += sellQuantity;
```
retotrinkler: think we should use ds-math here

src/Fund.sol
```diff
@@ -731,14 +732,15 @@ contract Fund is DSMath, DBC, Owned, Shares, FundInterface, ERC223ReceivingContr
...
+            if (exchanges[i].isApproveOnly) {
+                totalSellQuantityInApprove += sellQuantity;
             }
         }
         if (totalSellQuantity == 0) {
             isInOpenMakeOrder[sellAsset] = false;
         }
-        return totalSellQuantity;
+        return totalSellQuantity - totalSellQuantityInApprove; // Since quantity in approve is not actually in custody
```
retotrinkler: think we should use ds-math here

PR #429, opened 2018-03-23
Update Asset and Exchange structs

travs:


GauthamN:
On fund level, right now it doesn't check if the exchange is whitelisted and doesn't obtain adapter from canonical register (I might have missed it). Or will be added in another PR?

travs:
@GauthamN Yes that is in another PR. Will actually close this one and open with the appropriate branch as soon as possible. Think this PR may have been a bit premature; my mistake

PR #446, opened 2018-05-31
Change competition terms and conditions hash, change CHF asset mock a…

GauthamN:
…ddress

Copy/paste from CHANGELOG.md is good

### Added
- NoComplianceCompetition contract
- Deployment settings for competition-replica on kovan

### Changed
- T and C hash in competition contract
- T and C hash to bytes instead of bytes32

### Fixed
- Rounding error in tests

### DIFF
--- a/src/compliance/NoComplianceCompetition.sol
+++ b/src/compliance/NoComplianceCompetition.sol
@@ -1,0 +1,24 @@
+pragma solidity ^0.4.19;
+
+import "./NoCompliance.sol";
+
+/// @title Compliance Contract
+/// @author Melonport AG <team@melonport.com>
+/// @notice Only allow manager to invest in the Fund.
+contract NoComplianceCompetition is NoCompliance {
+
+    // PUBLIC VIEW METHODS
+
+    /// @notice Always returns true
+    /// @param x Address
+    /// @return True
+    function isCompetitionAllowed(
+        address x
+    )
+        view
+        returns (bool)
+    {
+        return true;
+    }
+
+}
--- a/src/compliance/NoComplianceCompetition.sol
+++ b/src/compliance/NoComplianceCompetition.sol
@@ -4,7 +4,7 @@ import "./NoCompliance.sol";
 
 /// @title Compliance Contract
 /// @author Melonport AG <team@melonport.com>
-/// @notice Only allow manager to invest in the Fund.
+/// @notice Anyone is allowed to invest / redeem
 contract NoComplianceCompetition is NoCompliance {
 
     // PUBLIC VIEW METHODS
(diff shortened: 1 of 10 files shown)


utils/deploy/contracts.js
```diff
@@ -138,20 +138,25 @@ async function deployEnvironment(environment) {
...
-        ethTokenAddress, deployed.CanonicalPriceFeed.address, deployed.OnlyManagerCompetition.address
+        ethTokenAddress, deployed.CanonicalPriceFeed.address, complianceAddress
       ],
       () => {}, true
     );
     deployed.FundRanking = await deployContract("FundRanking", opts);
     const blockchainTime = await getChainTime();
-    deployed.Competition = await deployContract("competitions/Competition", opts, [mlnAddr, chfAddress, deployed.Version.address, accounts[0], blockchainTime, blockchainTime + 864000, 22 * 10 ** 18, 10 ** 24, 1000]);
-    await deployed.Competition.instance.batchAddToWhitelist.postTransaction(opts, [10 ** 25, [accounts[0]]]);
+    deployed.Competition = await deployContract("competitions/Competition", opts, [mlnAddr, chfAddress, deployed.Version.address, accounts[0], blockchainTime, blockchainTime + 864000, 20 * 10 ** 18, 10 ** 24, 1000]);
```
jennazenk: could we add after this a transfer of MLN tokens to the Competition contract ?

src/compliance/NoComplianceCompetition.sol
```diff
@@ -0,0 +1,24 @@
+pragma solidity ^0.4.19;
+
+import "./NoCompliance.sol";
+
+/// @title Compliance Contract
+/// @author Melonport AG <team@melonport.com>
+/// @notice Only allow manager to invest in the Fund.
```
jennazenk: outdated comment, would be "Anyone is allowed to invest in the Fund". 

What is the difference between NoComplianceCompetition and NoCompliance ?

GauthamN: Changed the comment. It's basically the same with the added isCompetitionAllowed function that always returns true (actual version checks for whitelist and competition is active conditions)

utils/deploy/contracts.js
GauthamN: Yeah but could we assume the deployer has MLN tokens?

jennazenk: think we can but better to check with @travs

src/compliance/NoComplianceCompetition.sol
jennazenk: gotcha

PR #517, opened 2018-10-19
Move all code into /legacy & transactionFactory

schmidsi:
# Move to /legacy
I also removed the old deploy command since you probably want to reimplement them with TypeScript anyways.

Old integration tests still run.

# Transaction Factory
Have a look at: https://github.com/melonproject/smart-contracts/blob/0b25c4cb3925df2efb9684bbe7b64d4d88277663/src/contracts/factory/transactions/createComponents.ts
And https://github.com/melonproject/smart-contracts/blob/0b25c4cb3925df2efb9684bbe7b64d4d88277663/src/contracts/factory/transactions/continueCreation.ts

For maximum and minimum usage of `transactionFactory`

There is an issue for the migration for the rest. Actually an easy but tedious one: https://github.com/melonproject/smart-contracts/issues/519


travs:
This looks good; I need to read more about generics all the same

PR #528, opened 2018-10-26
Add engine

travs:
Add melon engine contract and some preliminary tests


schmidsi:
> Need to add a couple things; don't merge this yet @schmidsi

👍 Just tell me when it's ready

PR #529, opened 2018-10-26
Build

schmidsi:
With this pull-request, it is now easily possible to publish alpha versions with: `yarn publish:alpha`.

The published version can be used in every system, that supports ES modules (node, webpack, ...). Browser bundling does not work at the moment, but we don't need this currently anyways: The front-end does not depend on the protocol any more (only the GraphQL server).

One highlight is that the consuming application can deploy the contracts to their dev-chain with:

```bash
yarn melon deploy
```

More info in the README. Read the source, Luke ;)


.gitignore
```diff
@@ -32,5 +27,5 @@ coverage/
 # ignore cache
 .cache/
 
-# ignore concatenated files (only updated at verification time)
-utils/etherscan/concatenated
+# Doxity files ??
+legacy/scripts/
```
travs: yeah we will can even just get rid of these I think

legacy/contracts/risk-management/AssetList.sol
```diff
@@ -3,68 +3,37 @@ pragma solidity ^0.4.21;
 import "../dependencies/Owned.sol";
 
 // Generic AssetList
+// should this inherit from Policy?
```
travs: don't think so since it does not implement `rule` or `position`; looks like it is just used as a list in a couple other rules

PR #536, opened 2018-11-08
Add engine

travs:
Made a new branch for this.
Add melon engine contract, integration with fund, and some preliminary tests.


src/contracts/engine/transactions/sellAndBurnMln.test.ts
```diff
@@ -0,0 +1,217 @@
...
+
+const shared: any = {};
+
+// TODO: use the actual sellAndBurnMln typescript function rather than testing like this
+beforeAll(async () => {
+  await initTestEnvironment();
+  shared.env = getGlobalEnvironment();
+  shared.accounts = await shared.env.eth.getAccounts();
+  const wethAddress = await deployToken('ETH');
+  shared.mln = await deployAndGetContract('dependencies/token/BurnableToken', [
```
schmidsi: Hmm, I understand that it might be more time-consuming to write all calls and transactions first and it's just faster to directly get the web3 contract instance. Which is fine in the sense of rapid prototyping.

But we should aim for having all relevant calls and transactions as stand alone files. The `callFactory` and `transactionFactory` can be as easy as that:

```ts
import { callFactory } from '~/utils/solidity';
import { Contracts } from '~/Contracts';

const getHub = callFactory('getHub', Contracts.Spoke);

export { getHub };

// ---

import { transactionFactory } from '~/utils/solidity';
import { Contracts } from '~/Contracts';

export const continueCreation = transactionFactory(
  'continueCreation',
  Contracts.FundFactory,
);
```

The advantage of this model is that we can slowly refactor and enhance these functions then as we see need, like adding guards and post-processing, without changing the consumers at all or only a little.

What I want to say: Consider using the factories to quickly create simple method files which are easily extensible. But using the contracts directly in the tests is fine for now I think. But only in the tests. `getContract` & `deployAndGetContract` should'nt be exported at all.

src/contracts/engine/transactions/sellAndBurnMln.test.ts
```diff
@@ -0,0 +1,217 @@
...
+    shared.env.eth.sendTransaction({
+      from: shared.env.wallet.address,
+      to: shared.engine.options.address,
+      value: 1000000,
+    }),
+  ).rejects.toThrow('revert');
+});
+
+test('eth sent via contract selfdestruct is not tracked', async () => {
+  const sendEth = 10 ** 18;
```
schmidsi: Be carefull with these numbers. `10 ** 18` is already greater than `Number.MAX_SAFE_INTEGER (9007199254740991)`. Better to use `token-math/big-integer`.

src/contracts/engine/transactions/sellAndBurnMln.ts
```diff
@@ -0,0 +1,70 @@
...
+  prepareTransaction,
+  sendTransaction,
+  getContract,
+} from '~/utils/solidity';
+import { getToken } from '~/contracts/dependencies/token';
+import { isAddress } from '~/utils/checks';
+import { ensure } from '~/utils/guards';
+import { Contracts } from '~/Contracts';
+
+const guards = async (
```
schmidsi: Better to use the transactionFactory. Sorry, I know it is quite new, but it makes this stuff much easier.

src/utils/solidity/getContractWithPath.ts
```diff
@@ -0,0 +1,24 @@
...
+import { Address } from '~/utils/types';
+import { getGlobalEnvironment, Environment } from '~/utils/environment';
+
+export type GetContractWithPathFunction = (
+  relativePath: string,
+  address: Address,
+  environment?: Environment,
+) => typeof Eth.Contract;
+
+export const getContractWithPath: GetContractWithPathFunction = R.memoizeWith(
```
schmidsi: Whats the difference to `getContract`? The possibility to not have to use `Contracts` enum? Why not just add these contracts to the enum then?

src/utils/solidity/increaseTime.ts
```diff
@@ -0,0 +1,35 @@
+import { getGlobalEnvironment, Environment } from '../environment';
+
+export const increaseTime = async (
```
schmidsi: 👍

tslint.json
```diff
@@ -2,7 +2,7 @@
   "extends": ["tslint-config-airbnb", "tslint-config-prettier"],
   "rules": {
     "interface-name": false,
-    "max-line-length": [true, 80],
+    "max-line-length": [false],
```
schmidsi: max-line-length really helps for readability. Use `// tslint:disable-next-line:max-line-length` to disable it specifically

legacy/tests/integration/fundTrading.js
```diff
@@ -55,13 +55,22 @@ test.before(async t => {
...
+    [false],
+    deployed.TestingPriceFeed.options.address
   ).send({from: manager, gasPrice: config.gasPrice});
   await deployed.FundFactory.methods.continueCreation().send({from: manager, gasPrice: config.gasPrice});
   await deployed.FundFactory.methods.setupFund().send({from: manager, gasPrice: config.gasPrice});
   const fundId = await deployed.FundFactory.methods.getLastFundId().call();
   const hubAddress = await deployed.FundFactory.methods.getFundById(fundId).call();
   fund = await getFundComponents(hubAddress);
+  await Promise.all(Object.values(fund).map(async (component) => {
+    await deployed.MockVersion.methods.setIsFund(component.options.address).send({from: manager});
```
schmidsi: This sets isFund for all fund components? Why?

package.json
```diff
@@ -22,14 +22,15 @@
     "lint:eslint": "eslint .",
     "lint:tslint": "tsc --noEmit && tslint src/**/*.ts",
     "lint:prettier": "yarn prettier --list-different src/**/*.ts",
+    "deploy:typescript": "yarn ts-node -r tsconfig-paths/register -r dotenv/config src/utils/deploySystem.ts",
```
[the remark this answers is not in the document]
travs: Ok we can get rid of this.

src/contracts/engine/transactions/sellAndBurnMln.test.ts
```diff
@@ -0,0 +1,217 @@
...
+  getContract,
+} from '~/utils/solidity';
+import { Contracts } from '~/Contracts';
+
+const shared: any = {};
+
+// TODO: use the actual sellAndBurnMln typescript function rather than testing like this
+beforeAll(async () => {
+  await initTestEnvironment();
+  shared.env = getGlobalEnvironment();
```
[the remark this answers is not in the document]
travs: ok that is good!

src/contracts/engine/transactions/sellAndBurnMln.test.ts
```diff
@@ -0,0 +1,217 @@
+import { BigNumber } from 'bignumber.js';
```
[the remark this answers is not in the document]
travs: Ok yeah, we had some BigInteger import errors a couple weeks ago that we couldn't resolve at the end of a day I think, so will need to go back over this and convert to BigInteger

src/contracts/engine/transactions/sellAndBurnMln.test.ts
travs: Fair; goes along with using BigInteger above instead of this

src/contracts/engine/transactions/sellAndBurnMln.test.ts
travs: Ok yea that sounds good to me :)

src/utils/solidity/getContractWithPath.ts
travs: Yeah I think I was just trying to avoid polluting the enum with all the testing contracts but there's no harm in doing that really. We can factor out getContractWithPath if we want then

tslint.json
travs: This was the error where the linting tools were fighting each other. I will try it again re-enabled and see if it works.

legacy/tests/integration/fundTrading.js
travs: The `isFund` query on `MockVersion` is dumb and only tracks everything in a mapping. When we make the actual version contract we will have to change it a bit so that components (which are not funds themselves) can pay amgu. May also need a rename at that point. Anyway yeah this is just there to pass an assertion when paying amgu in tests at the moment

PR #523, opened 2018-10-23
Develop into refactor

travs:
Merge commits from develop into refactor

PR #540, opened 2018-11-12
Refactor

travs:
Just moves `refactor` in `develop`
Supersedes #523


package.json
```diff
@@ -108,6 +108,7 @@
     "commander": "^2.19.0",
     "debug": "^4.0.1",
     "dotenv": "^6.0.0",
+    "emoji-regex": "^7.0.1",
```
jennazenk: Gautham couldn't compile without this, any idea why @schmidsi

schmidsi: Actually no. This is very strange and shouldn't happen.

PR #548, opened 2018-11-15
Redeem

jennazenk:
### Added
- redeem
- redeemQuantity
- getFundHoldings

### Changed
- transactionFactory and prepare to allow for not estimating gas when needed


src/utils/solidity/prepareTransaction.ts
```diff
@@ -15,11 +15,12 @@ export interface PreparedTransaction {
 
 export const prepareTransaction = async (
   transaction,
+  options?,
   environment = getGlobalEnvironment(),
 ): Promise<PreparedTransaction> => {
   const encoded = transaction.encodeABI();
 
-  if (transaction.gasEstimation === undefined) {
+  if (options === undefined || options.gas === undefined) {
```
schmidsi: Makes sense

PR #551, opened 2018-11-16
WIP: Amgu tests

schmidsi:
It basically works, but refactoring in transactionFactory lead to breaking tests: --> We need to manually parse the logs in order to have an external signer (ethers, ledger, metamask, etc) supported.

I have it already in my head how to do it, but I think I need to sync with @fubhy closer with the refactoring of the transacitonFactory.


src/utils/solidity/transactionFactory.ts
```diff
@@ -136,25 +164,58 @@ const transactionFactory: TransactionFactory = <Args, Result>(
...
+
     transaction.name = name;
     const prepared = await prepareTransaction(
       transaction,
-      options,
+      optionsOrCallback,
       environment,
     );
-    return { ...prepared, contract };
+    const options: Options =
```
schmidsi: @fubhy I had to change a lot here: 1. Add in AMGU calc, 2. return a JSON serializable transaction to be able to pipe it through graphql to the frontend and back.

src/utils/solidity/transactionFactory.ts
```diff
@@ -164,13 +225,49 @@ const transactionFactory: TransactionFactory = <Args, Result>(
...
+      .sendTransaction(prepared.rawTransaction)
+      .then(null, error => {
+        throw new Error(
+          `Transaction failed for ${name}(${prepared.transactionArgs.join(
+            ', ',
+          )}): ${error.message}`,
+        );
+      });
+
+    const events = receipt.logs.reduce((carry, log) => {
```
schmidsi: Manual transaction log parsing. This is needed because we loose the web3 transaction object that does this automatically for us through GraphQL and hardware wallets.

Anyways, our own solution is now able to parse all events that we have in our contracts, not only the one in the current ABI.

src/contracts/engine/AmguConsumer.sol
```diff
@@ -23,7 +23,7 @@ contract AmguConsumer is DSMath {
         uint ethToPay = mul(
             sub(initialGas, gasleft()),
             mul(mlnPerAmgu, ethPerMln)
-        );
+        ) / 1 ether;
```
travs: @jennazenk Be aware of this one. Was talking to Simon and it looks like this is correct; the amgu payments we were getting before had 18 extra decimal places from what we expected

src/contracts/factory/FundFactory.sol
```diff
@@ -103,7 +103,7 @@ contract FundFactory is AmguConsumer {
         address[] _defaultAssets,
         bool[] _takesCustody,
         address _priceSource
-    ) public step(1) amguPayable {
+    ) public payable step(1) amguPayable {
```
travs: yup exactly

src/utils/solidity/transactionFactory.ts
travs: Very cool!!

src/contracts/fund/participation/transactions/executeRequest.ts
```diff
@@ -1,46 +1,38 @@
...
-import { getRequest } from '../calls/getRequest';
-import { createQuantity, greaterThan } from '@melonproject/token-math/quantity';
-import { Contracts } from '~/Contracts';
-import { getToken } from '~/contracts/dependencies/token';
-import { getHub, getSettings, ensureIsNotShutDown } from '../../hub';
+import { executeRequestFor } from './executeRequestFor';
+import { getGlobalEnvironment } from '~/utils/environment';
+import { Address } from '@melonproject/token-math/address';
 
-const guard = async (params, contractAddress, environment) => {
```
jennazenk: why are we removing the guards here?

src/utils/solidity/prepareTransaction.ts
```diff
@@ -15,25 +33,60 @@ export interface PreparedTransaction {
...
+  const maxGasPrice = multiply(
+    toBI(defaultOptions.gasLimit),
+    toBI(defaultOptions.gasPrice),
+  );
+
+  const balance = await environment.eth.getBalance(options.from);
+  const maxAmguInEth = subtract(toBI(balance), maxGasPrice);
+
+  // We don't know the amgu price at this stage yet, so we just send all
+  // available ETH for the gasEstimation. This should throw if amgu price
```
jennazenk: can we find a way to know the amgu price before that ? Dont like the idea of sending the entire balance ; doesnt feel right

src/contracts/engine/AmguConsumer.sol
jennazenk: thanks

src/utils/solidity/prepareTransaction.ts
jennazenk: @schmidsi

src/contracts/fund/participation/transactions/executeRequest.ts
jennazenk: @schmidsi

schmidsi: This is now just a wrapper function around `executeRequestFor`, which does the guards check then. I ran into problems with AMGU forwarding, I have the following hypothesis:

Given the two functions:
```solidity
function executeRequestFor(address requestOwner) public payable amguPayable {  /* ... do it */ }
function executeRequest() public payable amguPayable {
    executeRequestFor(msg.sender);
}
```

Where executeRequest() just calls executeRequestFor(), the amgu calculation is made twice and that leads to errors. So the hypothesis is:

** An amguPayable function cannot call another amguPayable function **

Actually, I think we can remove `executeRequest` completely from the contracts. --> Contracts should have as little code as possible. Every line of code enhances the possibility for bugs.

jennazenk: ok @travs we should think of removing executeRequest as Simon said !

src/utils/solidity/prepareTransaction.ts
jennazenk: ok gotcha nice

src/contracts/fund/participation/transactions/executeRequest.ts
travs: @schmidsi yeah `executeRequest` was just for convenience but it is indeed a bit needless. I will remove it. AmguPayable was intentionally left off that function before this PR

PR #553, opened 2018-11-19
Ethfinex refactor

GauthamN:
- Ethfinex Adapter and associated thirdparty contracts
- Ethfinex Adapter handles wrapping, unwrapping tokens and also accounting for them


bin/compile.js
```diff
@@ -56,9 +56,9 @@ var writeFiles = function (compileOutput, contract) {
     debug('Writing', contract);
     mkdirp.sync(targetDir);
     if (fs.existsSync(targetBasePath + ".abi")) {
-        throw new Error(
-        // tslint:disable-next-line:max-line-length
-        "Contract name duplication detected: " + targetBasePath + ".abi. Please make sure that every contract is uniquely named inside the same directory.");
+        // throw new Error(
```
schmidsi: This can actually lead to weird problems. Which contracts are colliding? Can we rename one of them?

travs: re-generated and this was fixed it appears 08f0e66

Issue #546, opened 2018-11-14
Fund unit tests


travs:
Goal is to get back to the level of coverage we had in our previous unit tests.
This includes adding these tests:

## Participation

- [x] make request
- [x] execute request
- [x] cancel request
- [x] redeem

## Vault

- [x] withdraw

## Accounting

- [x] basic share price calculation
- [x] permissioned functions

## Fee manager

- [x] permissioned functions
- [x] basic fee rewarding

## Trading

- [x] basic trading state variables check, callback functionality check
- [x] returning tokens to vault

## Policies

- [x] each risk management rule
- [ ] each compliance rule

PR #554, opened 2018-11-21
Unit trading

GauthamN:
Trading unit tests
#546


src/contracts/fund/trading/unit/trading.test.ts
```diff
@@ -0,0 +1,124 @@
...
+      takesCustodyMasks,
+    ]),
+  ).rejects.toThrow(errorMessage);
+});
+
+test('returnToVault sends back token balances to the vault', async () => {
+  const tokenQuantity = new BigInteger(10 ** 20);
+
+  await shared.mln.methods
+    .transfer(shared.trading.options.address, `${tokenQuantity}`)
```
schmidsi: You could use the `contracts/dependencies/token/transactions/` here.

src/utils/constants/orderSignatures.ts
```diff
@@ -0,0 +1,61 @@
...
+
+function joinSignature(name, argTypes = []) {
+  return `${name}(${argTypes.join(',')})`;
+}
+
+function abiEncode(joinedSignature) {
+  return Abi.encodeFunctionSignature(joinedSignature);
+}
+
+export const makeOrderSignature = joinSignature('makeOrder', [
```
schmidsi: Actually I don't understand the value of this. We should derive these signatures directly from the ABI, or just provide strings. Why make an array and join it directly?

src/utils/helpers/randomAddress.ts
```diff
@@ -1,4 +1,5 @@
 import * as Web3Utils from 'web3-utils';
 import { Address } from '~/utils/types';
 
-export const randomAddress = () => new Address(Web3Utils.randomHex(20));
+export const randomAddress = () =>
+  new Address(Web3Utils.randomHex(20)).toLowerCase();
```
schmidsi: Why `toLowerCase`? I think its good practice to try to always use the check-summed addresses. But I also know that here and there that leads to problems (for example comparing).

GauthamN: Changed it back

src/utils/constants/orderSignatures.ts
GauthamN: We need signature string along with bytes. I didn't find web3 function that returns signature string by taking function name as parameter. And joining as a separate step is just for better readability of signature string.

PR #556, opened 2018-11-21
Invest unit

travs:
#546


src/contracts/prices/KyberPriceFeed.sol
```diff
@@ -9,7 +10,7 @@ import "../exchanges/thirdparty/kyber/KyberNetworkProxy.sol";
 /// @notice Where external data includes sharePrice of Melon funds
 /// @notice PriceFeed operator could be staked and sharePrice input validated on chain
 /// @notice TODO: ERC20Clone inconsistency
-contract KyberPriceFeed is SimplePriceFeed {
+contract KyberPriceFeed is PriceSourceInterface, DSThing {
```
GauthamN: Just to confirm,  DSThing for inherting ownership contract right?

travs:
@schmidsi yes

src/contracts/prices/KyberPriceFeed.sol
travs: yes, gives us auth (owned extension), math and note (unused).

PR #560, opened 2018-11-22
Trading OasisDex

jennazenk:
#### Calls
- getOasisDexOrder
- getFundOpenOrder
- getExchangeIndex
 - isOasisDexMakePermitted
- isOasisDexTakePermitted
- ensureFundNotShutDown
- ensureFundOwner
- ensureMakePermitted
- ensurePriceExistsOnAssetPair

#### Transactions
From an account 
- makeOrderFromAccountOasisDex
- takeOrderFromAccountOasisDex
- cancelOrderFromAccountOasisDex

From a fund 
- callOnExchange
- makeOasisDexOrder
- takeOasisDexOrder 
- cancelOasisDexOrder

### Changed
- event parser

NB: this PR also contains the deployment script for Kyber + the function swapTokensFromAccount, but that transaction still fails.


src/contracts/exchanges/transactions/deployKyberEnvironment.ts
```diff
@@ -0,0 +1,301 @@
...
+/* eslint no-bitwise: ["error", { "allow": ["&"] }] */
+function bytesToHex(byteArray) {
+  const strNum = Array.from(byteArray, (byte: any) =>
+    `0${(byte & 0xff).toString(16)}`.slice(-2),
+  ).join('');
+  const num = `0x${strNum}`;
+  return num;
+}
+
+export const deployKyberEnvironment = async (
```
schmidsi: Would be better to write this as an object:
```
{  deployer,
  mlnToken,
  ethToken,
  eurToken, }

File: src/contracts/fund/trading/calls/getExchangeIndex.ts [hunk 2]
```diff
@@ -0,0 +1,29 @@
+import { getContract } from '~/utils/solidity';
+import { Contracts } from '~/Contracts';
+import { ensure } from '~/utils/guards/ensure';
+
+const getExchangeIndex = async (
```
schmidsi: The general function signature is:
```ts
const fn = (contractAddress, paramsObject, environment)
```

Where as the `contractAddress` is always the address of the contract this function acts upon and `paramsObject` should always be an object.

So this signature should be:
```ts
const getExchangeIndex = async (tradingAddress, { exchangeAddress }, environment)
```

src/contracts/fund/trading/calls/getFundOpenOrder.ts
```diff
@@ -0,0 +1,30 @@
...
+    environment,
+  );
+  const order = await tradingContract.methods.orders(index).call();
+  const makerToken = await getToken(order.makerAsset);
+  const takerToken = await getToken(order.takerAsset);
+  return {
+    exchangeAddress: order.exchangeAddress,
+    id: web3Utils.toDecimal(order.orderId),
+    updateType: order.updateType,
+    makerAsset: order.makerAsset,
```
schmidsi: `makeAsset` and `takerAsset` are probably superfluous since the address of the `makerAsset` is already inside `makerQuantity`

src/contracts/fund/trading/calls/getExchangeIndex.ts
schmidsi: Usage of the `callFactory` would enforce this behaviour and also give added niceties like the possibility to subscribe to a call.

src/utils/solidity/transactionFactory.ts
```diff
@@ -15,9 +15,12 @@ import {
   Options,
 } from '../solidity';
 import { Address } from '../types';
+import { ensure } from '../guards';
 
-type TransactionArg = number | number[] | string | string[];
-type TransactionArgs = TransactionArg[];
+// TODO: Fix the types here once the transaction factory decorators are
```
schmidsi: This can lead to weird problems: If the EVM receives other stuff than strings/numbers, it fails with misleading error messages.

PR #574, opened 2018-11-28
0x make & take order from fund

schmidsi:
This is also a rather big pull request, so please merge soon.


package.json
```diff
@@ -31,37 +31,37 @@
...
+    "precompile": "rimraf out",
+    "publish:alpha": "yarn check && yarn publish --tag alpha --new-version $(semver $npm_package_version -i prerelease --preid alpha)",
+    "test": "JSON_RPC_ENDPOINT=ws://localhost:8545 yarn jest --runInBand --forceExit"
   },
   "devDependencies": {
-    "0x.js": "1.0.7",
+    "@0x/types": "^1.3.0",
     "@types/jest": "^23.3.9",
     "@types/node": "^10.12.10",
+    "babel-polyfill": "^6.26.0",
```
fubhy: I don't think we need that.

schmidsi: We do unfortunately: It seems like 0x project did not build their packages properly.

Issue #581, opened 2018-11-29
Compile only what has changed


travs:
Compiling takes around 2 mins on an i7 which limits how fast we can prototype and test the solidity code.
We could take advantage of caching to only compile the contracts that have changed and their dependants.

I know how to do this with `make` but it means maintaining a Makefile (not that big a deal if that's what we have to do), but maybe there is a better way that you guys know of. [`jake`](http://jakejs.com) looks like it may be useful, or maybe [this compiler](https://github.com/pubkey/solidity-cli) or [that one](https://0xproject.com/docs/sol-compiler#introduction), but I haven't had time to experiment with any of this.

As a side effect, we could also cache in CI and shave a few minutes off our builds.

 @schmidsi @fubhy Thoughts?

schmidsi:
Yes, I observed the same effect. Actually, after adding the interfaces. It seems to slow compilation down tremendously.

One big problem at the moment is that we recompile all contracts all the time. It is possible to only recompile a specific contract and its dependencies. In fact, I already started to implement this functionality here: https://github.com/melonproject/protocol/blob/a4ee46f78c8d9c803978008a6319b24c6bfb3ace/bin/compile.ts#L39:L68 but did not properly end it.

Solidity-cli looks interesting too.

I will play around a bit with these two solutions.

travs:
Just to clarify what I mean about the tree: if we make changes to just one contract, only that contract and its dependants (not dependencies) need to be recompiled.

Simple example:
```
"->" = depends on

A -> B
B -> C
```
If we change `B` here, we don't need to recompile `C`, but only `A` and `B`.

schmidsi:
Thanks for the clarification. The problem is, that I don't know yet, how to use `C`'s already compiled bytecode from cache, during recompilation of `B`. That's why we just have to recompile `C` when `B` is compilling.

To actually have the changes mirrored to `A`, one needs to compile everything, or explicitly target `A`:  `yarn compile path/to/A.sol`.

travs:
Compiling is sufficiently fast now that we got rid of some external contracts we were compiling with our own.

PR #584, opened 2018-11-30
Possibility to compile only one file

schmidsi:
For example, run:

```bash
DEBUG=* yarn compile src/contracts/engine/Engine.sol
```

That finishes in only 10s on my machine.

Probably closes: https://github.com/melonproject/protocol/issues/581

# How it works
If provided a glob to `yarn compile`, it compiles only the files the glob matches. The default glob is `src/contracts/**/*.sol`. It follows then all files for dependencies and resolves them.

A heavy dependent contract like EthfinexAdapter still takes a long time to compile though.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,0 +1,4 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.13.1
+ignore: {}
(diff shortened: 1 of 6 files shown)


travs:
Nice! Thanks Simon :)

PR #588, opened 2018-12-04
Clean up management fee code, remove gav variable

GauthamN:
Clean up management fee code, remove gav variable


travs:
makes sense!

Issue #587, opened 2018-12-03
Add isFund logic to Version


travs:
Instead of just `setIsFund` manually being called for each component, which was just a temporary measure.

PR #590, opened 2018-12-05
Version and ranking

### LINKED ISSUE #570: Version + Ranking contracts
Need to add back these contracts

travs:
Fixes #570


src/contracts/factory/FundRanking.sol
```diff
@@ -0,0 +1,29 @@
+pragma solidity ^0.4.21;
+pragma experimental ABIEncoderV2;
+
+import "./FundFactory.sol";
+import "../fund/accounting/Accounting.sol";
+
+contract FundRanking {
+    function getFundDetails(address _factory)
```
jennazenk: @gempi this is the function you'll need to call in melon.js in order to get the ranking

jennazenk:
> Fixes #570

Fixes #587 also

PR #600, opened 2018-12-06
Add Registrar

### LINKED ISSUE #573: Pricefeeds to inherit registrar?
We need a way to register assets and allowed exchange function signatures by some registrar. If we include this functionality in the pricefeeds, this means all the pricefeeds must inherit registrar interface

### LINKED ISSUE #591: Possibility to add/remove assets in Melon asset universe + possibility to add/remove exchanges
at version level I think; what do you think @GauthamN @travs

travs:
Fixes #591 
Fixes #573


src/contracts/version/transactions/registerAsset.ts
```diff
@@ -0,0 +1,54 @@
+import web3EthAbi from 'web3-eth-abi';
+import { Address } from '@melonproject/token-math/address';
+import {
+  transactionFactory,
+  PrepareArgsFunction,
+  EnhancedExecute,
+} from '~/utils/solidity/transactionFactory';
+import { Contracts } from '~/Contracts';
+import { FunctionSignatures } from '~/contracts/fund/trading/utils/FunctionSignatures';
+interface RegisterAssetArgs {
```
travs: Main thing we can change here is to pass in a `token-math#Token` instance and derive some of this stuff.
This is fine for the time being but just noting

src/contracts/version/transactions/registerExchange.ts
```diff
@@ -0,0 +1,39 @@
...
+import { Address } from '@melonproject/token-math/address';
+import {
+  transactionFactory,
+  PrepareArgsFunction,
+  EnhancedExecute,
+} from '~/utils/solidity/transactionFactory';
+import { Contracts } from '~/Contracts';
+import { FunctionSignatures } from '~/contracts/fund/trading/utils/FunctionSignatures';
+interface RegisterExchangeArgs {
+  exchange: Address;
```
travs: Similar to [above comment](https://github.com/melonproject/protocol/pull/600/files#r239409022), we may be able to pass in `ExchangeConfig` instances similar to how we do in `createComponent` to achieve some code reuse.

Issue #561, opened 2018-11-23
Proposal to flatten contracts folder structure


schmidsi:
We have now a quite deep & nested folder structure in the contracts folder with no clear (or at least not for me) organisation.

So I propose a new layout:

From:
```
src/contracts/
├── dependencies
│   └── token
├── engine
├── exchanges
│   ├── thirdparty
│   ├── 0x
│   │   ├── ethfinex
│   │   ├── kyber
│   │   └── oasisdex
├── factory
├── fund
│   ├── accounting
│   ├── compliance
│   ├── fees
│   ├── hub
│   ├── participation
│   ├── policies
│   ├── risk-management
│   ├── shares
│   ├── trading
│   └── vault
├── prices
├── testing
└── version
```
To:
```
├── dependencies
│   ├── token
|   ├── dapphub
│   ├── 0x
│   ├── ethfinex
│   ├── kyber
│   └── oasisdex
├── engine
├── factory
├── accounting
├── fees
├── hub
├── participation
├── policies
├── shares
├── trading
├── vault
├── prices
├── testing
└── version
```

Explanation:
- All thirdparty code / dependencies are isolated in the `dependencies` folder. Especially the code for external exchanges should be moved there.
- The former `exchanges` folder gets merged with the `trading` folder. Exchange adapters can be seen as a utility for the trading functionality.
- `compliance` and `risk-management` is merged into `policies`
- `fund` is flattened out. The folder structure does not need to represent the dependency structure. If so, everything should be a subfolder of `version`, then `factory` and so on.

Thoughts?

GauthamN:
I agree with first point

travs:
would classify this as lower priority but yeah most of it sounds good.
would keep `fund/` since it's pretty plain that it holds fund components, not sure what you mean about dependency structure since it's not like that now

schmidsi:
Sure, low prio.

Regarding `fund/`. I understand the intention. My thinking is: The flatter, the better.

jennazenk:
>would keep fund/ since it's pretty plain that it holds fund components, not sure what you mean about dependency structure since it's not like that now

Agree to that. Also would keep compliance and risk management separated. Otherwise sounds good, we can do this when we are back on schedule.

GauthamN:
Can be closed?

travs:
@GauthamN Yes

PR #611, opened 2018-12-07
WIP: Flat imports out

schmidsi:
This PR rewrites the solidity source file resolver in a way, that we can omit the relative paths:

So instead of:
```solidity
import "../dependencies/token/ERC20.i.sol";
```

We write:
```solidity
import "ERC20.i.sol";
```

This has the advantage that we could reorganize our folder structure without changing the hash of the actual contract source files as described in #561. Furthermore, a future split into several repositories would only require to change the resolver logic, also without changing the hash of the contract files.

The disadvantages are: It could break tooling. For example my VS code extension shows an error (File import callback not supported). Solium does not complain about this.


schmidsi:
Tests not passing yet. I'll investigate more on Monday.

schmidsi:
But it is already meaningful to discuss this idea and have a look at the solidity source changes

travs:
Nice! When using dapp we had this behaviour only for the dapp modules, but I like the idea for everything else as well.

GauthamN:
I had an issue earlier with duplicate contract names when we had a flat compilation structure. What happened is, a contract with the same name in a thirdparty directory overwrote another one after compilation. But i guess we have a check for this now? Just something to keep in mind security-wise. Thoughts @schmidsi @travs

schmidsi:
> I had an issue earlier with duplicate contract names when we had a flat compilation structure. What happened is, a contract with the same name in a thirdparty directory overwrote another one after compilation. But i guess we have a check for this now? Just something to keep in mind security-wise. Thoughts @schmidsi @travs

I ensure that this situation will throw an error. Thanks for the heads-up.

src/contracts/fund/policies/compliance/UserWhitelist.sol
```diff
@@ -1,17 +1,17 @@
...
-contract Whitelist is Policy, DSAuth {
+contract UserWhitelist is Policy, DSAuth {
 
     event ListAddition(address indexed who);
     event ListRemoval(address indexed who);
 
     mapping (address => bool) whitelisted;
 
-    function Whitelist(address[] _preApproved) public {
+    function UserWhitelist(address[] _preApproved) public {
```
travs: nice

src/contracts/exchanges/thirdparty/0x/Exchange.sol
```diff
@@ -19,9 +19,9 @@
 pragma solidity ^0.4.21;
 pragma experimental ABIEncoderV2;
 
-import "../../../dependencies/token/StandardToken.sol";
-import "./LibOrder.sol";
-import "./Ownable.sol";
+import "StandardToken.sol";
+import "LibOrder.sol";
+import "Ownable.sol";
```
travs: maybe we should just rename this file to ZeroExExchange, since importing `Exchange.sol` is not totally clear

schmidsi: Yeah, I thought about that too. But ABI files get named after the actual contract name, not the file name holding the contract. But I think we should not change the actual contract names of 3rd party contracts.

travs: Yeah fair enough

PR #623, opened 2018-12-13
Re-organize upper levels of contracts

### LINKED ISSUE #615: Register funds in registry
The funds should be registered *across* versions, since the engine must be able to tell who is a fund without iterating a storage array of versions.

travs:
Now the `Registry` has more responsibility, and funds are registered there rather than only on the Version.
Amgu price is now set on the Engine itself, instead of the Version (because Version changes).
Version securityShutDown disables new fund creation. 

Registry tracks the funds as mentioned above, but also now tracks the Versions, and furthermore the PriceSource and the Engine (which might never change, but still).

Now to set up you have to do:
```
beginSetup(...)
createAccounting(environment, version);
createFeeManager(environment, version);
createParticipation(environment, version);
createPolicyManager(environment, version);
createShares(environment, version);
createTrading(environment, version);
createVault(environment, version);
completeSetup(environment, version);
```

Note that first step does not require user to pay amgu, since that causes a stack too deep error in solidity. May be a way around it in the future, but for now I think it's fine to pay at all other steps.

Fixes #615 

@john-0x Some changes were made to the contracts that will affect docs


jennazenk:
@fubhy @gempi fyi fund setup has been modified in that PR

PR #624, opened 2018-12-13
Debug ethfinex

luongnt95:


src/Contracts.ts
```diff
@@ -152,8 +152,8 @@ export const requireMap = {
     require('../out/version/MockVersion.abi.json'),
   [Contracts.WrapperLock]:
     require(`../out/${Contracts.WrapperLock}.abi.json`),
-  [Contracts.WrapperlockEth]:
-    require(`../out/${Contracts.WrapperlockEth}.abi.json`),
+  [Contracts.WrapperLock]:
```
schmidsi: Should be `WrapperlockEth`, no?

luongnt95: you are right

PR #634, opened 2018-12-15
Do todos

travs:
- remove calcSharePriceAndAllocateFees (2 separate steps now)
- move triggerRewardAllFees to Accounting
- remove UserBlackList
- remove redundant functions in Accounting (getFundHoldingsLength, calcUnclaimedFees)
- reduce max held assets to 20
- update initialization logic in spoke
- check that exchange and exchange methods are allowed before trading
- use hub.routes instead of hub.settings


jennazenk:
@john-0x special alert here - now if there is no valid price, an investor can redeem by bypassing the fees

PR #637, opened 2018-12-17
Fix shareprice & add getNativeToken

gempi:
- Fix shareprice call & add getNativeToken

### DIFF
--- a/src/contracts/fund/accounting/calls/performCalculations.ts
+++ b/src/contracts/fund/accounting/calls/performCalculations.ts
@@ -5,11 +5,15 @@ import {
   createQuantity,
   QuantityInterface,
 } from '@melonproject/token-math/quantity';
+import { getHub } from '~/contracts/fund/hub/calls/getHub';
+import { getSettings } from '../../hub/calls/getSettings';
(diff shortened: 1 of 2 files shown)


src/contracts/fund/accounting/calls/performCalculations.ts
```diff
@@ -18,10 +22,17 @@ const postProcess = async (
...
   const quoteToken = await getQuoteToken(environment, prepared.contractAddress);
+  const hub = await getHub(environment, prepared.contractAddress);
+  const settings = await getSettings(environment, hub);
+  const fundToken = await getToken(environment, settings.sharesAddress);
+
   const calculations = {
     gav: createQuantity(quoteToken, result.gav),
     nav: createQuantity(quoteToken, result.nav),
-    sharePrice: createQuantity(quoteToken, result.sharePrice),
+    sharePrice: getPrice(
```
jennazenk: hey @gempi , i don't think we need to call getPrice here; performCalculations already returns the price of the fund's share. What's the rationale?

jennazenk: sorry , Simon just informed me this is token-maths getPrice, not pricefeed getPrice!

PR #641, opened 2018-12-19
Add calls and make parallel

travs:
@fubhy 
Got rid of the stepFor function since we can do the creation steps in parallel now.

To see whether beginSetup is done, look at `managersToHubs(_guy)`. It's zero if you didn't do it yet.

The best way to check whether a component was created is by doing `FundFactory#managersToSettings(_guy)` and looking at the values for whatever key you are curious about.
Don't have the .ts for that .sol function yet.

To see if the `setupComplete` function was called, check `childExists(_hubAddress)`. Note that this is not using the manager's address, but the hub's (which you can get from `managersToHubs(_guy)`).

Also we renamed `Accounting` `quoteAsset` to `denominationAsset`, so I changed that in the ranking contract FYI.


schmidsi:
@gempi

PR #662, opened 2019-01-03
Fix exchange config

jennazenk:
### Changed

- Fix takesCustody in exchangeConfig
- Add makeOasisDexOrder, cancelOasisDexOrder and shutDownFund to playground-kyber tests


src/tests/system/playground-kyber.systest.ts
```diff
@@ -36,7 +41,7 @@ describe('playground', () => {
 
     const manager = withNewAccount(master);
     const trader = withNewAccount(master);
-
+    console.log('Manager address is: ', manager.wallet.address);
```
schmidsi: I suggest to use `log.debug` to have it in the log files too and is not swallowed by jest.

src/tests/utils/setupInvestedTestFund.ts
```diff
@@ -86,7 +86,7 @@ const setupInvestedTestFund = async (environment: Environment) => {
 
   debug('Testfund setup and invested', fundName, routes);
 
-  return routes;
+  return { ...routes, hubAddress };
```
schmidsi: 👍

PR #669, opened 2019-01-08
Fix naming

travs:
- version names must be unique
- if you register a fund with a name in one version, that name is yours alone to use in any version
- fund names can only contain `[a-zA-Z0-9 .-_]` (space included)

### DIFF
--- a/src/contracts/version/Registry.sol
+++ b/src/contracts/version/Registry.sol
@@ -70,6 +70,7 @@ contract Registry is DSAuth {
 
     mapping (address => address) public fundsToVersions;
     mapping (bytes32 => bool) public versionNameExists;
+    mapping (bytes32 => address) public fundNameHashToOwner;
 
     uint public constant MAX_REGISTERED_ENTITIES = 20;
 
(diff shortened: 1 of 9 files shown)


jennazenk:
@john-0x

src/contracts/version/Registry.sol
```diff
@@ -82,12 +85,46 @@ contract Registry is DSAuth {
...
+        bytes memory b = bytes(_name);
+        if (b.length > MAX_FUND_NAME_BYTES) return false;
+        for (uint i; i < b.length; i++){
+            bytes1 char = b[i];
+            if(
+                !(char >= 0x30 && char <= 0x39) && // 9-0
+                !(char >= 0x41 && char <= 0x5A) && // A-Z
+                !(char >= 0x61 && char <= 0x7A) && // a-z
+                !(char == 0x20 || char == 0x2D) && // space, dash
+                !(char == 0x2E || char == 0x5F)    // period, underscore
```
jennazenk: can we add also `|| char  == 0x2A` for *  (line 99)  @travs

travs: 0650ae883b71407f437f51a6ae2b77426daa614d

Issue #672, opened 2019-01-09
JSON Docs


schmidsi:
One of the most powerful tools to help other developers (internal and external) use our functions is to proper document them with JSDocs. We should document all important functions properly with JSDocs annotation. This seems like a tedious task but here is a simple guideline how to gradually improve our documentation coverage:

Every time one doesn't understand what a function does and goes into code to read about, they should just add some JSDocs to that function and write down the findings.

Here is a screenshot illustrating how to write JSDocs and how useful they are:

<img width="1552" alt="screen shot 2019-01-09 at 11 18 07" src="https://user-images.githubusercontent.com/103280/50895118-71507100-1405-11e9-9233-8058cb490854.png">

A more deep dive into JSDocs can be found here:
http://usejsdoc.org/index.html#block-tags

Note: It is superfluous to annotate type information in JSDocs because TypeScript is already doing it.

PR #675, opened 2019-01-09
Document approve as JSDoc example

schmidsi:
#672


jennazenk:
@john-0x this is a good example

PR #678, opened 2019-01-10
Fix names js

gempi:
- Fix names in js (Related to #669)


src/contracts/factory/calls/getFundDetails.ts
```diff
@@ -43,7 +44,7 @@ export const getFundDetails = async (
     return {
       address,
       creationTime: new Date(creationTimes[index] * 1000),
-      name: names[index],
+      name: web3Utils.hexToAscii(names[index]),
```
schmidsi: Trim this result: `web3Utils.hexToAscii(names[index]).trim()`

gempi:
@schmidsi please check again

PR #687, opened 2019-01-15
Track historical investors

travs:


jennazenk:
@john-0x for documentatipn
@iherger for monitoring tools

PR #690, opened 2019-01-16
Add takeOrderOnKyber

luongnt95:


src/contracts/fund/trading/transactions/takeOrderOnKyber.ts
```diff
@@ -0,0 +1,90 @@
...
+} from '~/utils/solidity/transactionFactory';
+import { getHub } from '../../hub/calls/getHub';
+import { getRoutes } from '../../hub/calls/getRoutes';
+import { ensureSufficientBalance } from '~/contracts/dependencies/token/guards/ensureSufficientBalance';
+import { ensureFundOwner } from '~/contracts/fund/trading/guards/ensureFundOwner';
+import { ensureIsNotShutDown } from '~/contracts/fund/hub/guards/ensureIsNotShutDown';
+import { ensureKyberTakePermitted } from '../guards/ensureKyberTakePermitted';
+import { FunctionSignatures } from '../utils/FunctionSignatures';
+
+// The order needs to be signed by the manager
```
jennazenk: can probably remove that comment

src/contracts/fund/trading/transactions/takeOrderOnKyber.ts
```diff
@@ -0,0 +1,90 @@
...
+const prepareArgs: PrepareArgsFunction<TakeOrderOnKyberArgs> = async (
+  environment,
+  { makerQuantity, takerQuantity, fillTakerQuantity = takerQuantity },
+  contractAddress,
+) => {
+  const exchangeIndex = await getExchangeIndex(environment, contractAddress, {
+    exchange: Exchanges.KyberNetwork,
+  });
+
+  const takerAddress = takerQuantity.token.address;
```
jennazenk: think this should `takerAddress` should be `takerAsset`  and `makerAddress` should be `makerAsset`

src/contracts/fund/trading/calls/isKyberTakePermitted.ts
```diff
@@ -0,0 +1,69 @@
+import web3EthAbi from 'web3-eth-abi';
+import { getContract } from '~/utils/solidity/getContract';
+import { Contracts, Exchanges } from '~/Contracts';
+import { getRoutes } from '~/contracts/fund/hub/calls/getRoutes';
+import { getHub } from '~/contracts/fund/hub/calls/getHub';
+import { FunctionSignatures } from '../utils/FunctionSignatures';
+import { Environment } from '~/utils/environment/Environment';
+import { Address, QuantityInterface } from '@melonproject/token-math';
+
+const isKyberTakePermitted = async (
```
schmidsi: I'm not 100% sure, but this looks to me like a "duplicated code" smell. What is the actual difference between `isKyberTakePermitted` and `isOasisDexMakePermitted`. I only see fillTakerQuantity which could be defaulted to `0`.

So better to refactor this function to `isTakePermitted` that works for all exchanges. Like the contracts are implemented if I understand them right.

https://github.com/melonproject/protocol/blob/develop/src/contracts/fund/policies/risk-management/PriceTolerance.sol#L73:L79

src/contracts/fund/trading/guards/ensureKyberTakePermitted.ts
```diff
@@ -0,0 +1,24 @@
+import { ensure } from '~/utils/guards/ensure';
+import { isKyberTakePermitted } from '../calls/isKyberTakePermitted';
+import { Environment } from '~/utils/environment/Environment';
+import { Address, QuantityInterface } from '@melonproject/token-math';
+
+const ensureKyberTakePermitted = async (
```
schmidsi: Same here. Could be just: `ensureTakePermitted`

src/contracts/fund/trading/calls/isKyberTakePermitted.ts
luongnt95: https://github.com/melonproject/protocol/blob/develop/src/contracts/fund/trading/calls/isOasisDexTakePermitted.ts#L47

schmidsi: Right. Please use the Exchange enum here: https://github.com/melonproject/protocol/blob/develop/src/Contracts.ts#L269

PR #702, opened 2019-01-16
DO NOT MERGE: Fee problem isolation

schmidsi:
During integration of fees in the front-end, @gempi found an interesting "effect". I'm not sure if it is a bug of the protocol or a bug of ganache or intended behaviour at all.

# Fee default values
We found that with fees enabled, the share price of a fund decreases automatically. It decreased very fast, so the first problem is probably the fee setup:

Currently in the tests we setup the fees like this:
```typescript
const fees = [
    {
      feeAddress: melonContracts.fees.managementFee.toLowerCase(),
      feePeriod: toBI(0),
      feeRate: appendDecimals(weth, 0.02),
    },
    {
      feeAddress: melonContracts.fees.performanceFee.toLowerCase(),
      feePeriod: toBI(DAY_IN_SECONDS * 90),
      feeRate: appendDecimals(weth, 0.2),
    },
  ];
```

It seems a bit weird to me to have a `feePeriod` of 0 for the management fee. What does this mean? Shouldn't it be something like one year or also 90 days? Should the manager be able to set this value during setup or shouldn't we provide just a sane default?

# Fee calculation on ganache.

[The test file that I wrote (`fee.test.ts`) to isolate this behaviour calculates the share price 3 times with a pause between them of 5 seconds](https://github.com/melonproject/protocol/compare/fee-problem-isolation?expand=1#diff-30aa8f3177d67c5123e20bcf056e432fR58). Now the sharePrice declines as follows:

800000000000000000
799999999594114663
799999999188229325

Interestingly, there are no new blocks between the two calls (as seen in the debug output attached). Both blocks have the same timestamp.

So the second question is: Shouldn't fee calculation only depend on the block timestamp? And if so, why do we see different results on the same block with the same timestamp but 5 seconds later?

This behaviour would also explain, why the other fee tests irrationally failed: Because if I set the pause to only 1 second, the effect is not visible.

# The problem in the frontend
Lastly, in the front-end we observed that there is a different value for ranking and for fund detail view. This is probably because both queries have different caching timeouts (@fubhy  ?)

```json
{
  "initialCalculations": {
    "gav": {
      "token": {
        "address": "0x8cc1a5B95d82f5b9a770D16fDF121aD4099795b9",
        "decimals": 18,
        "symbol": "WETH"
      },
      "quantity": "1000000000000000000"
    },
    "nav": {
      "token": {
        "address": "0x8cc1a5B95d82f5b9a770D16fDF121aD4099795b9",
        "decimals": 18,
        "symbol": "WETH"
      },
      "quantity": "799999999594114664"
    },
    "sharePrice": {
      "base": {
        "token": {
          "address": "0x48b7bD588f0D71A3b534a95Dcd180f0302C87072",
          "decimals": 18,
          "symbol": "MLNF"
        },
        "quantity": "1000000000000000000"
      },
      "quote": {
        "token": {
          "address": "0x8cc1a5B95d82f5b9a770D16fDF121aD4099795b9",
          "decimals": 18,
          "symbol": "WETH"
        },
        "quantity": "799999999594114663"
      }
    }
  },
  "ranking": [
    {
      "address": "0xB58E65da7284A2672379B9Dc24FC6C25a33Ca0d5",
      "creationTime": "2019-03-15T18:20:01.000Z",
      "name": "test-fund-6h29",
      "rank": 1,
      "sharePrice": {
        "base": {
          "token": {
            "decimals": 18,
            "symbol": "MLNF"
          },
          "quantity": "1000000000000000000"
        },
        "quote": {
          "token": {
            "address": "0x8cc1a5B95d82f5b9a770D16fDF121aD4099795b9",
            "decimals": 18,
            "symbol": "WETH"
          },
          "quantity": "799999999594114663"
        }
      }
    }
  ],
  "blockBefore": {
    "number": 4698,
    "hash": "0x9dcb089331c5c1bc90b95f16020574f6ad06fe9c7359f3d7d68d542e4ce3941d",
    "parentHash": "0x36f0fc48d6975cb21006ad245dcdbcb4ce05faba183c29181e1260d6ef14c5aa",
    "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "nonce": "0x0000000000000000",
    "sha3Uncles": "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
    "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000800000000000000000004000000000000000000000000000000000000000100000000000000000000000000040000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000000000000000000000000",
    "transactionsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
    "stateRoot": "0x35b2d46172a074f4dfaa05cb838f88269327acdb6780c9146b3566c7031f3a08",
    "receiptsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
    "miner": "0x0000000000000000000000000000000000000000",
    "difficulty": "0",
    "totalDifficulty": "0",
    "extraData": "0x",
    "size": 1000,
    "gasLimit": 8000000,
    "gasUsed": 150292,
    "timestamp": 1552674019,
    "transactions": [
      "0xdc6b3256d94ae57599aca1c3753c23d7819760b439f07dcb9acbb70df84b167f"
    ],
    "uncles": []
  },
  "blockAfter": {
    "number": 4698,
    "hash": "0x9dcb089331c5c1bc90b95f16020574f6ad06fe9c7359f3d7d68d542e4ce3941d",
    "parentHash": "0x36f0fc48d6975cb21006ad245dcdbcb4ce05faba183c29181e1260d6ef14c5aa",
    "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "nonce": "0x0000000000000000",
    "sha3Uncles": "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
    "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000800000000000000000004000000000000000000000000000000000000000100000000000000000000000000040000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000000000000000000000000",
    "transactionsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
    "stateRoot": "0x35b2d46172a074f4dfaa05cb838f88269327acdb6780c9146b3566c7031f3a08",
    "receiptsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
    "miner": "0x0000000000000000000000000000000000000000",
    "difficulty": "0",
    "totalDifficulty": "0",
    "extraData": "0x",
    "size": 1000,
    "gasLimit": 8000000,
    "gasUsed": 150292,
    "timestamp": 1552674019,
    "transactions": [
      "0xdc6b3256d94ae57599aca1c3753c23d7819760b439f07dcb9acbb70df84b167f"
    ],
    "uncles": []
  },
  "afterCalcs": {
    "gav": {
      "token": {
        "address": "0x8cc1a5B95d82f5b9a770D16fDF121aD4099795b9",
        "decimals": 18,
        "symbol": "WETH"
      },
      "quantity": "1000000000000000000"
    },
    "nav": {
      "token": {
        "address": "0x8cc1a5B95d82f5b9a770D16fDF121aD4099795b9",
        "decimals": 18,
        "symbol": "WETH"
      },
      "quantity": "799999999188229326"
    },
    "sharePrice": {
      "base": {
        "token": {
          "address": "0x48b7bD588f0D71A3b534a95Dcd180f0302C87072",
          "decimals": 18,
          "symbol": "MLNF"
        },
        "quantity": "1000000000000000000"
      },
      "quote": {
        "token": {
          "address": "0x8cc1a5B95d82f5b9a770D16fDF121aD4099795b9",
          "decimals": 18,
          "symbol": "WETH"
        },
        "quantity": "799999999188229325"
      }
    }
  }
}
```

### DIFF
--- a/src/contracts/dependencies/token/calls/balanceOf.test.ts
+++ b/src/contracts/dependencies/token/calls/balanceOf.test.ts
@@ -9,10 +9,10 @@ import { balanceOf } from '../calls/balanceOf';
 import { getToken } from '../calls/getToken';
 import { deployToken } from '../transactions/deploy';
 import { transfer } from '../transactions/transfer';
+import { delay } from 'rxjs/operators';
 
 describe('balanceOf', () => {
   const shared: any = {};
-  const delay = ms => new Promise(resolve => setTimeout(resolve, ms));
 
   beforeAll(async () => {
     shared.env = await initTestEnvironment();
--- a/src/contracts/factory/calls/getFundDetails.ts
+++ b/src/contracts/factory/calls/getFundDetails.ts
@@ -4,11 +4,12 @@ import { Address, createPrice, createQuantity } from '@melonproject/token-math';
 import { getContract } from '~/utils/solidity/getContract';
 import { Contracts } from '~/Contracts';
 import { getTokenByAddress } from '~/utils/environment/getTokenByAddress';
+import { Environment } from '~/utils/environment/Environment';
 
 export const getFundDetails = async (
-  environment,
-  contractAddress: Address,
-  versionAddress: Address,
+  environment: Environment,
+  contractAddress: Address = environment.deployment.melonContracts.ranking,
+  versionAddress: Address = environment.deployment.melonContracts.version,
 ) => {
   const contract = getContract(
     environment,
--- a/src/contracts/fund/trading/transactions/cancelOasisDexOrder.ts
+++ b/src/contracts/fund/trading/transactions/cancelOasisDexOrder.ts
@@ -14,10 +14,12 @@ import { Exchanges, Contracts } from '~/Contracts';
 import { FunctionSignatures } from '../utils/FunctionSignatures';
 import { emptyAddress } from '~/utils/constants/emptyAddress';
 
-export type CancelOasisDexOrderResult = any;
+export interface CancelOasisDexOrderResult {
+  id: string;
+}
 
 export interface CancelOasisDexOrderArgs {
-  id?: string;
+  id: string;
   maker: Address;
   makerAsset: Address;
   takerAsset: Address;
--- a/src/index.ts
+++ b/src/index.ts
@@ -69,3 +69,6 @@ export {
   makeOasisDexOrder,
 } from '~/contracts/fund/trading/transactions/makeOasisDexOrder';
 export { withDifferentAccount } from '~/utils/environment/withDifferentAccount';
+export {
+  cancelOasisDexOrder,
+} from '~/contracts/fund/trading/transactions/cancelOasisDexOrder';
--- a/src/tests/integration/fee.test.ts
+++ b/src/tests/integration/fee.test.ts
@@ -1,0 +1,89 @@
+import { allLogsWritten } from '../utils/testLogger';
+import { Environment } from '~/utils/environment/Environment';
+import { deployAndInitTestEnv } from '../utils/deployAndInitTestEnv';
+import { setupInvestedTestFund } from '../utils/setupInvestedTestFund';
+import { createPrice, createQuantity } from '@melonproject/token-math';
+import { getTokenBySymbol } from '~/utils/environment/getTokenBySymbol';
+import { update } from '~/contracts/prices/transactions/update';
+import { getLogCurried } from '~/utils/environment/getLogCurried';
+import { performCalculations } from '~/contracts/fund/accounting/calls/performCalculations';
+import { delay } from 'rxjs/operators';
+import { getFundDetails } from '~/contracts/factory/calls/getFundDetails';
+import { getLatestBlock } from '~/utils/evm';
+
+const getLog = getLogCurried('melon:protocol:systemTest:playground');
+
+describe('feeTests', () => {
+  const shared: {
+    env?: Environment;
+    [p: string]: any;
+  } = {};
+
+  beforeAll(async () => {
+    shared.env = await deployAndInitTestEnv();
+    shared.accounts = await shared.env.eth.getAccounts();
+    shared.routes = await setupInvestedTestFund(shared.env);
+
+    const weth = getTokenBySymbol(shared.env, 'WETH');
+    const mln = getTokenBySymbol(shared.env, 'MLN');
+
+    const mlnPrice = createPrice(
+      createQuantity(mln, '1'),
+      createQuantity(weth, '2'),
+    );
+
+    const ethPrice = createPrice(
+      createQuantity(weth, '1'),
+      createQuantity(weth, '1'),
+    );
+
+    await update(shared.env, shared.env.deployment.melonContracts.priceSource, [
+      ethPrice,
+      mlnPrice,
+    ]);
+  });
+
+  afterAll(async () => {
+    await allLogsWritten();
+  });
+
+  test('', async () => {
+    const log = getLog(shared.env);
+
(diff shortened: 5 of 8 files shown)


jennazenk:
Thanks for the report! 

First, regarding feePeriod setup is correct. Management fee period is always 1 year (hardcoded in the contract), so the 0 passed doesn't matter. Performance fee period is at the discretion of the manager, what you passed as 3 months is correct as well. 

The share price decreasing over time is expected behavior; the reason for that is because the share price returned by `performCalculations` is net of management and performance fees. This means that this share price is calculated using the (total existing number of shares + feeShares not yet created but due to the manager in exchange for management/performance). See here: https://github.com/melonproject/protocol/blob/develop/src/contracts/fund/accounting/Accounting.sol#L130

The second thing to note, is that fees are calculated by seconds, which explains different values on the same block but at different time (e.g. 5 seconds apart). 

The third thing to note (and maybe most important here) is that there are different "share prices". 
- **Redemption share price**: This is the share price returned by `performCalculations` under the sharePrice key and `calcSharePrice`. This share price is net of all fees (management + performance). This is the price at which an investor can redeem his shares. 
- **Investment share price**: This share price is returned by `performCalculations` under the `gavPerShareNetManagementFee` key (see [here](https://github.com/melonproject/protocol/blob/develop/src/contracts/fund/accounting/Accounting.sol#L134)) . This is the share price at which you can invest in the fund - ie. this is the share price that should be displayed in the investment form. This share price is only net of management fee (you don't pay fee for past management), but not net of performance fees (you pay performance fee for the whole period, even if you enter at the beginning of the period). There needs to be a proper disclaimer at investment step that clearly indicates that if you invest after the beginning of a period, you will pay for the performance of the whole period. Recommendation is to invest in a fund at the beginning of a period. 

We need to agree on when to display `sharePrice` (redemption price) and when to display `gavPerShareNetManagementFee` (investment price) on the frontend. 

@gempi @schmidsi @GauthamN

schmidsi:
Thanks for clarification.

I understood that the share price is declining over time if holdings and prices stay the same.

What confused me is the difference of the calculation on the same block with the same block number (see the attached logs). As far as I know does the EVM not know the current time, only the block time and [that is used for the calculations](https://github.com/melonproject/protocol/blob/develop/src/contracts/fund/fees/ManagementFee.sol#L24).

But it makes sense that the EVM does not know the block timestamp of now, because the current block is not mined yet. Therefore, it uses probably the current time for the calls to these view functions [as described here](https://forum.ethereum.org/discussion/2452/block-timestamp-a-k-a-now-in-constant-functions).

PR #706, opened 2019-01-17
Frontend trade integration

schmidsi:


schmidsi:
Merging bymyself to not block front-end dev. Nothing interesting to see here ;)

PR #712, opened 2019-01-18
Add kyber pricefeed updater script

GauthamN:


bin/melon.js
```diff
@@ -271,6 +271,68 @@ program
...
+        updateKyber,
+      } = require('../lib/contracts/prices/transactions/updateKyber');
+
+      try {
+        await updateKyber(environment, environment.deployment.melonContracts.priceSource);
+      } catch (err) {
+        console.error(err);
+      }
+
+      setTimeout(
```
schmidsi: I think this should be `setInterval`

PR #713, opened 2019-01-18
Add updater to Kyber pricefeed

GauthamN:


jennazenk:
@travs please comment ok when you reviewed

PR #719, opened 2019-01-21
Fix getActiveOasisDexOrders

schmidsi:
I probably need some smart-contract help for this one: It seems that OasisDex orders made from fund are not takeable from account.

Furthermore, they do not show up in `getActiveOasisDexOrders`. Orders from account show up as seen in: https://github.com/melonproject/protocol/blob/develop/src/tests/integration/accountTrading.test.ts#L95-L96

I'm not 100% sure if it is truly a problem in the fund trading or if `getActiveOasisDexOrders` does something wrong AND I do not make a mistake by calling `takeOrderFromAccountOasisDex`. But so far, I could not find any other test that tests this: make order from fund -> take that order from account. Only make order from fund -> cancel order.

The order does not get matched by MatchingMarket since these tests deploy an empty one.

Next step would probably be to directly check `offers` for that particular order.

### DIFF
--- a/src/contracts/exchanges/third-party/oasisdex/MatchingMarket.sol
+++ b/src/contracts/exchanges/third-party/oasisdex/MatchingMarket.sol
@@ -72,13 +72,13 @@ contract SimpleMarket is EventfulMarket, DSMath {
     }
 
     modifier can_buy(uint id) {
-        require(isActive(id), "Order not active");
+        require(isActive(id));
         _;
     }
 
     modifier can_cancel(uint id) {
-        require(isActive(id), "Order not active");
-        require(getOwner(id) == msg.sender, "Order not owned by sender");
+        require(isActive(id));
+        require(getOwner(id) == msg.sender);
         _;
     }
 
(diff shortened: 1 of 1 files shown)


jennazenk:
I am able to make an order from a fund and take it from an account; you can check it in playground-oasisdex. https://github.com/melonproject/protocol/pull/721

src/contracts/exchanges/MatchingMarketAccessor.sol
```diff
@@ -16,30 +17,34 @@ contract MatchingMarketAccessor {
...
+        for (uint i = 0; i < 1000; i++) {
+            if (id == 0) {
+                break;
+            }
+
             if (market.isActive(id)) {
                 address sellGem;
                 address buyGem;
-                (, sellGem, , buyGem) = market.getOffer(ids[i]);
+                (, sellGem, , buyGem) = market.getOffer(id);
```
fubhy: This was the problem @jennazenk

schmidsi:
Ok, the error was not in `takeOrderFromAccountOasisDex` but in `getActiveOasisDexOrders`. So this branch is good to merge now.

jennazenk:
only problem is I think we shouldn't change the MatchingMarket contract no?

schmidsi:
> only problem is I think we shouldn't change the MatchingMarket contract no?

Yeah, I mean it was only adding revert messages to troubleshoot which does not actually change the ABI or behaviour of the contract. I'm fine of rejecting these commits if there are serious security concerns.

PR #727, opened 2019-01-24
Tests for multiple investors in same fund + investment in different assets 

jennazenk:
- requestedShares argument is not optional anymore, it is required (in requestInvestment and invest) 

@schmidsi @gempi


src/contracts/fund/hub/transactions/setupFund.ts
```diff
@@ -0,0 +1,84 @@
+import * as R from 'ramda';
```
schmidsi: I think this is a test-only file? Can you please move it to `src/tests/utils`? There are already other files like this.

src/contracts/fund/participation/transactions/executeRequestFor.ts
```diff
@@ -30,9 +31,7 @@ const guard = async (environment, params, contractAddress) => {
     'Amount of requested shares is null',
   );
 
-  // TODO: remaining pre flights
-  // ensure isRecent
-  // ensure isPriceRecent
+  await ensureHasValidRequest(environment, contractAddress, params.who);
```
schmidsi: 👍

src/contracts/fund/participation/transactions/invest.ts
```diff
@@ -0,0 +1,31 @@
+import { getRoutes } from '~/contracts/fund/hub/calls/getRoutes';
```
schmidsi: I think this is a test-only file? Can you please move it to src/tests/utils? There are already other files like this.

src/contracts/fund/hub/transactions/setupFund.ts
jennazenk: no its not a test file ; it allows to create a fund without investing in it

src/contracts/fund/participation/transactions/invest.ts
jennazenk: same as setupFund

src/contracts/fund/hub/transactions/setupFund.ts
jennazenk: Can be used by bots / Midas as a way to easily spin up a fund

schmidsi: True.

PR #750, opened 2019-02-05
Mainnet deploy

schmidsi:


src/contracts/factory/transactions/completeSetup.ts
```diff
@@ -33,5 +33,5 @@ export const completeSetup = transactionFactory(
   undefined,
   undefined,
   postProcess,
-  { amguPayable: true },
+  { amguPayable: true, gas: '7500000' },
```
schmidsi: I think we should be able to remove these. The problem was somewhere else ...

PR #762, opened 2019-02-13
Update oasisdex exchange address

jennazenk:
I did not make a new release but just updated the address


schmidsi:
Build fails but it seems like it is just one of those "normal" hickups. I restart them.

PR #835, opened 2019-12-11
Feature/add callonexchange orderdata array

SeanJCasey:
Refactors `bytes makerAssetData` and `bytes takerAssetData` into a `bytes[4] orderData` parameter in `Trading().callOnExchange()`. The extra slots in the array are reserved for `makerFeeAssetData` and `takerFeeAssetData` that are introduced in 0x v3.

Note:
- As `indexed` and `bytes[4]` don't work well together in solidity 0.4.x, I needed to remove `indexed` from two params of `event ExchangeMethodCall`. These need to be restored when we update to solidity 0.5.x (@luongnt95 )


SeanJCasey:
Closing to rebase this against solidity 0.5.x branch

PR #860, opened 2019-12-30
fix: remove defaultAssets from ownedAssets at fund creation

### LINKED ISSUE #823: Remove code that adds defaultAssets to ownedAssets at fund creation
From @SeanJCasey

The `defaultAssets` passed in when setting up a Fund are used in multiple places.
It is useful to pass these in for a manager to accept investments immediately in arbitrary assets (this is in `Participation.sol`).

It probably doesn't make sense to use it [in `Accounting.sol`](https://github.com/melonproject/protocol/blob/master/src/contracts/fund/accounting/Accounting.sol#L37-L39) though, where we add each to ownedAssets, while owning none of each.

This should be an easy change, but we just have to be careful.

EDIT: we could also rename the variable to `defaultInvestmentAssets` if we go through with this change

luongnt95:
Fixes #823


src/factory/FundFactory.sol
```diff
@@ -172,7 +171,7 @@ contract FundFactory is AmguConsumer, Factory {
     {
         managersToRoutes[msg.sender].participation = participationFactory.createInstance(
             managersToHubs[msg.sender],
-            managersToSettings[msg.sender].defaultAssets,
+            managersToSettings[msg.sender]._defaultInvestmentAssets,
```
travs: is this being accessed properly (should drop the `_` I think)? @luongnt95

travs:
Fixes #823

src/factory/FundFactory.sol
luongnt95: yes and fixed.

PR #863, opened 2020-01-02
Use docker to compile in CI

### LINKED ISSUE #862: Integrate docker into this repo
We can use docker to bring in dependencies and execute scripts/programs in a cross-platform way, and have the same environment as our CI will have.

This issue is fixed when we have docker integrated and the CI build passes with it.

[Relevant](https://github.com/avantgardefinance/box/blob/ef26bf7a5fed3e042c1ed32cad2438fd0a39e494/docker-compose.yml).

CC: @fubhy

travs:
fixes #862

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 DIR = ./compile
+SOLC_COMMAND = docker run --rm -i ethereum/solc:0.5.15 solc --standard-json
 files =
+
 all: clean extract thirdparty_contracts
 thirdparty_contracts: extract
 	cp thirdparty/* out/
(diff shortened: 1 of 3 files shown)


fubhy:
@travs can we also merge this into the other branch for Ishak to be able to build on his mac :-P

Makefile
```diff
@@ -1,5 +1,7 @@
 DIR = ./compile
+SOLC_COMMAND = docker run --rm -i ethereum/solc:0.5.15 solc --standard-json
```
fubhy: How does this even work without the volume mounting?

fubhy: Aaaah, right, I remember, you are actually bundling all the contract code into JSON before. Makes sense. Then this works of course.

PR #865, opened 2020-01-03
fix: MaxConcentration to evaluate post-trade concentration

### LINKED ISSUE #821: Max concentration needs to account for value being traded for
As of now, the max concentration rule only accounts for an asset's current GAV as a percentage of total GAV:

https://github.com/melonproject/protocol/blob/ca2d3b0d41e0cb936ce7428e32515dddd2b59f40/src/contracts/fund/policies/risk-management/MaxConcentration.sol#L33

It should actually account for what the estimated asset GAV should be after the proposed trade has been executed.

travs:
fixes #821 

Rebase onto, and merge after #870

### DIFF
--- a/src/fund/policies/risk-management/MaxConcentration.sol
+++ b/src/fund/policies/risk-management/MaxConcentration.sol
@@ -5,6 +5,7 @@ import "../../../prices/IPriceSource.sol";
 import "../../accounting/Accounting.sol";
 import "../../trading/Trading.sol";
 import "../TradingSignatures.sol";
+import "../../../prices/IPriceSource.sol";
 
 contract MaxConcentration is TradingSignatures, DSMath {
     enum Applied { pre, post }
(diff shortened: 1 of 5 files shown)


src/fund/policies/risk-management/MaxConcentration.sol
```diff
@@ -20,19 +21,46 @@ contract MaxConcentration is TradingSignatures, DSMath {
...
+        address takerToken = (sig == TAKE_ORDER) ? addresses[2] : addresses[3];
+        if (denominationAsset == takerToken) { return true; }
+
+        uint concentration;
+        uint totalGav = accounting.calcGav();
+        if (sig == MAKE_ORDER) {
+            IPriceSource priceSource = IPriceSource(Hub(Trading(msg.sender).hub()).priceSource());
+            address makerToken = addresses[2];
+            uint makerQuantiyBeingTraded = values[0];
+            address takerToken = addresses[3];
```
travs: somewhat duplicate of line 35; can remove line 35 and put in `else` block here as well

PR #870, opened 2020-01-08
Upgrade to solidity 0.6.1

### LINKED ISSUE #849: Upgrade to solidity 0.6.0
[Breaking changes from 0.5.x](https://solidity.readthedocs.io/en/latest/060-breaking-changes.html).

Also fix compiler warnings where applicable (should be fewer now that ABIEncoderV2 is out of experimental).

travs:
fixes #849

### DIFF
--- a/src/exchanges/KyberAdapter.sol
+++ b/src/exchanges/KyberAdapter.sol
@@ -1,4 +1,4 @@
-pragma solidity 0.5.15;
+pragma solidity 0.6.1;
 pragma experimental ABIEncoderV2;
 
 import "../dependencies/WETH.sol";
@@ -37,7 +37,7 @@ contract KyberAdapter is DSMath, ExchangeAdapter {
         bytes[4] memory orderData,
         bytes32 identifier,
         bytes memory signature
-    ) public onlyManager notShutDown {
+    ) public override onlyManager notShutDown {
         Hub hub = getHub();
 
         require(
@@ -72,7 +72,7 @@ contract KyberAdapter is DSMath, ExchangeAdapter {
             targetExchange,
             bytes32(0),
             Trading.UpdateType.take,
-            [address(uint160(makerAsset)), address(uint160(takerAsset))],
+            [payable(makerAsset), payable(takerAsset)],
             [actualReceiveAmount, takerAssetAmount, takerAssetAmount]
         );
     }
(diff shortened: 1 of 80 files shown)


SeanJCasey:
@travs One further change we can make here is to replace the `address(uint160(myAddress))` pattern with `payable(myAddress)`

tests/integration/fundMaliciousToken.test.js
```diff
@@ -143,6 +142,8 @@ describe('fund-malicious-token', () => {
       .requestInvestment(dummyAmount, dummyAmount, maliciousToken.options.address)
       .send({ ...investorTxOpts, value: amguAmount });
 
+    await increaseTime(5); // to avoid executing in same block as update
```
SeanJCasey: This is fine here, but just FYI, I'm introducing a simple `delay(ms)` helper in `fix/deploy-testnet` that forces a real pause, so I will refactor this to use that helper once merged. This will allow our tests that only need to increment time by a small amount (several seconds) to work on testnets, since they don't rely on evm manipulation.

tests/unit/exchanges/ethfinexOrder.test.js
```diff
@@ -161,7 +161,12 @@ test('Previously made ethfinex order cancelled and not takeable anymore', async
 });
 
 test('Withdraw (unwrap) maker asset of cancelled order', async () => {
-  increaseTime(25*60*60);
+  // skip past order expiration
+  const depositLockEnd = await zrxWrapperLock.methods.depositLock(
+    trading.options.address
+  ).call();
+  const latestBlockTimestamp = (await web3.eth.getBlock()).timestamp;
+  await increaseTime(depositLockEnd - latestBlockTimestamp + 1);
```
SeanJCasey: Good catch! It is definitely best to be intentional with the time intervals we set. It's tough to debug when there are no comments about why we're waiting a particular amount of time.

src/exchanges/KyberAdapter.sol
```diff
@@ -185,7 +185,7 @@ contract KyberAdapter is DSMath, ExchangeAdapter {
     /// @param srcToken Address of src token
     /// @param destToken Address of dest token
     /// @param srcAmount Amount of src token
-    /// @return destAmount Amount of dest token expected in return
+    /// @param destAmount Amount of dest token expected in return
```
SeanJCasey: This param is already defined in line 186. The `@return` should be for `minRate`

PR #879, opened 2020-01-14
fix: better amgu assertion

### LINKED ISSUE #866: Better assertions for `amguPayable` functions in tests
Right now, the `amgu.test.js` integration tests assert whether the actual amount of ETH charged for an `amguPayable` function is less than an estimated amount. The estimated amount calculation more-or-less mirrors the `amguConsumer` modifier code:

https://github.com/melonproject/protocol/blob/5780305a10b86a888a180b85e55957f57e13f1c7/tests/integration/amgu.test.js#L39

This needs to be more precise than greater than / less than, either:
A. [ideal] exact amounts
B. some tolerated % deviation from the estimate

Also, it would be nice to implement a `calcAmguPayableInEther` helper, which could be used by `amguPayable` and also for external estimations like our tests, for consistency...

luongnt95:
fixes #866

### DIFF
--- a/src/engine/AmguConsumer.sol
+++ b/src/engine/AmguConsumer.sol
@@ -16,15 +16,19 @@ abstract contract AmguConsumer is DSMath {
     function mlnToken() public view virtual returns (address);
     function priceSource() public view virtual returns (address);
     function registry() public view virtual returns (address);
+    event Gas(uint preGas, uint postGas);
 
     /// bool deductIncentive is used when sending extra eth beyond amgu
     modifier amguPayable(bool deductIncentive) {
-        uint initialGas = gasleft();
+        uint preGas = gasleft();
         _;
+        uint postGas = gasleft();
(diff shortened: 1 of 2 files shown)


SeanJCasey:
Nice job thinking of exposing the gas consumed via an event to web3 @luongnt95 ! This will be helpful for testing.

Can we make the event a bit more of an "event" and give it a more specific namespacing?

E.g., maybe in `AmguConsumer`, we can define an `AmguPaid` event like this, which will be emitted at the end of the `amguPayable` modifier:

```
event AmguPaid(unit256 _amguFeePaid, uint256 _amguChargableGas, uint256 _incentivePaid);
```

`_amguFeePaid` = The ETH amount of Amgu paid
`_amguChargableGas` = The amount of gas used to calculate `_amguPaid`
`_incentivePaid` = The incentive fee paid

PR #930, opened 2020-02-20
Change network id for local testnet.

fubhy:
I'd like to also change this for the current v1.1.0 tracking branch (doesn't exist currently) so we can use that in our development environment (box) if possible. Any objections or thoughts?


fubhy:
Such a branch could be use to cherry-pick upstream (develop branch) changes that don't affect the contract code but e.g. improve developer ergonomics or so for us in the frontend.

travs:
We can chat about this in keybase, but my understanding was that `master` was always tracking the latest release. Good idea re: cherry-picking non-contract fixes from other branches

PR #933, opened 2020-02-21
Feature/call on exchange encoded data param

luongnt95:


travs:
Deprecated in favour of #955

PR #955, opened 2020-03-06
Feature/integrate airswap

### LINKED ISSUE #973: Make CallOnExchange abstract
This is just tracking the effort to take any parameters into CallOnExchange and decode them.

### LINKED ISSUE #974: Integrate with Airswap
Tracking the by now basically-complete task of integrating with Airswap.

luongnt95:
Fixes #973, fixes #974


src/fund/trading/Trading.sol
```diff
@@ -168,10 +154,34 @@ contract Trading is TokenUser, Spoke, TradingSignatures {
...
+    {
+        if (_methodSelector == TAKE_ORDER) {
+            address adapter = exchanges[_exchangeIndex].adapter;
+            (bool success, bytes memory returnData) = adapter.delegatecall(
+                abi.encodeWithSelector(
+                    OrderTaker(adapter).extractTakeOrderRiskManagementArgs.selector,
+                    _encodedArgs
+                )
+            );
+            require(success, "Encoded arguments might not match");
```
travs: see PR-level comment about how we might move this into OrderTaker (or OrderFIller)

tests/integration/fund0xV2Trading.test.js
```diff
@@ -127,37 +128,40 @@ describe('Fund takes an order', () => {
...
+    orderValues[2] = signedOrder.makerFee;
+    orderValues[3] = signedOrder.takerFee;
+    orderValues[4] = signedOrder.expirationTimeSeconds;
+    orderValues[5] = signedOrder.salt;
+    orderValues[6] = fillQuantity;
+    orderData[0] =  signedOrder.makerAssetData;
+    orderData[1] = signedOrder.takerAssetData;
+
+    const hex = web3.eth.abi.encodeParameters(
+      ['address[4]', 'uint256[7]', 'bytes[2]', 'bytes'],
```
travs: it might make sense to maintain a js mapping (in constants.js maybe) from `Integration` to ``EncodingTypes`. Maybe we can also expose this as a required variable on each adapter (see PR-level comment), and we can get the signature to encode by dynamically as we need it.

src/exchanges/adapters/ZeroExV2Adapter.sol
```diff
@@ -9,72 +9,85 @@ import "../libs/OrderTaker.sol";
...
+            bytes[2] memory orderData,
+        ) = __decodeTakeOrderArgs(_encodedArgs);
+
+        riskManagementAddresses = [
+            orderAddresses[0],
+            orderAddresses[1],
+            __getAssetAddress(orderData[0]),
+            __getAssetAddress(orderData[1]),
+            address(0),
+            address(0)
```
SeanJCasey: Why aren't there maker/taker fee addresses for 0x v2? Shouldn't it ZRX, at least when there is a fee?

src/exchanges/adapters/ZeroExV2Adapter.sol
```diff
@@ -131,47 +146,32 @@ contract ZeroExV2Adapter is ExchangeAdapter, OrderTaker {
...
+
+        IRegistry registry = __getRegistry();
+        require(registry.assetIsRegistered(
+            __getAssetAddress(orderData[0])), 'Maker asset not registered'
         );
-        require(
-            __getAssetAddress(_orderData[1]) == _orderAddresses[3],
-            "__validateTakeOrderParams: takerAssetData does not match address"
+        require(registry.assetIsRegistered(
+            __getAssetAddress(orderData[1])), 'Taker asset not registered'
```
SeanJCasey: It doesn't matter too much, but @travs and I started to move to this pattern of "functionName: reason" for revert strings, e.g., "__validateTakeOrderParams: takerAssetData does not match address". It's not a huge deal, but you can apply this to this PR as well if you'd like.

src/exchanges/adapters/ZeroExV2Adapter.sol
```diff
@@ -230,4 +230,46 @@ contract ZeroExV2Adapter is ExchangeAdapter, OrderTaker {
...
+    function __decodeTakeOrderArgs(
+        bytes memory _encodedArgs
+    )
+        public
+        pure
+        returns (
+            address[4] memory orderAddresses,
+            uint256[7] memory orderValues,
+            bytes[2] memory orderData,
+            bytes memory signature
```
SeanJCasey: For both 0x adapters, it might be nice to make `uint256 takerFillAmount` its own decoded value, as it doesn't refer to the 0x order, but is the desired fill amount for taking this order. As mentioned above, we could also have `uint256 makerFillAmount` (however named), so that we can perform extra validation, but that would be optional / not strictly-necessary.

Maybe even re-name the params `zeroExOrderAddresses`, etc, so it's clear that these params are specific to the 0x order, not the Melon order.

src/exchanges/libs/OrderTaker.sol
```diff
@@ -59,37 +71,38 @@ abstract contract OrderTaker is OrderFiller {
...
+        );
     }
 
     // INTERNAL FUNCTIONS
- 
+
     /// @notice Reserved function for executing a take order on an external exchange
     /// @dev When executing your order, use the values in __fillAssets and __fillExpectedAmounts
     /// @dev Include the `validateAndFinalizeFilledOrder` modifier
     /// @param _fillData Encoded data used by the OrderFiller
```
SeanJCasey: Could add natspec here and below for `@param _encodedArgs`, just to explain what these are in the file.

SeanJCasey:
@luongnt95 Can you please add a unit test suite for the Airswap adapter? See the other in this folder: https://github.com/melonproject/protocol/tree/feature/integrate-airswap/tests/unit/exchanges/adapters

PR #965, opened 2020-03-21
[WIP] Feature: add/remove multiple exchange adapters

### LINKED ISSUE #910: Add method to remove exchange(s) and add multiple exchanges on Trading
We have a method to add exchanges after fund setup, but no method to remove.
We can also add methods for each which add/remove multiple exchanges in one tx.

cc: @iherger

SeanJCasey:
Fixes #910 . It also moves towards the nomenclature that will be used when we start abstracting exchanges into "Integrations".

- update `addExchange` to `enableAdapters`, which accepts multiple adapters
- add `disableAdapters`
- change state to use an `EnumerableSet`, which allows us to get rid of the concept of `exchangeIndex`. Frontends are currently required to know the exchange's position in the `exchanges` array, but now `callOnExchange` is called with an exchange adapter address. More importantly, it also limits the size of the `enabledAdapters` (formerly `exchanges`) array, so that adding and removing the same adapters maintains a predictable size, disallowing an infinitely large array.
- updates namespaces and removes the redundant `getExchangeInfo()`

(The build and tests are currently expected to fail, as it does not alter the code for `callOnExchange` or `validateCallOnExchange`, since these change substantially in #933 . This PR will need to be rebased on top of that one, change those functions to remove the concept of exchangeIndex, then fix the tests that use the old `addExchange()` and `getExchangeInfo()`


SeanJCasey:
Closing this PR as it has been incorporated in and superseded by #977

PR #966, opened 2020-03-23
Refactor: interfaces

travs:
A few notes on the interfaces:

- they're minimal: just what is needed for inter-contract communication inside our own system
- avoided interface-to-interface inheritance even though it's possible, because it gets very messy when dealing with functions that return custom structs as we often use
- contracts should inherit their interface where possible, so we can enforce that these functions exist and don't end up with mysterious errors in testing due to unimplemented functions

One downside of not using interface-interface inheritance is that we need to use multiple interfaces to interact with a contract sometimes. For example, if we want to do `totalSupply` on Shares we would actually do `IERC20(myShares).totalSupply`, which is not the most intuitive.
Thankfully this naturally goes away the more we merge components.
We could also change this behaviour somehow else (e.g. implement functions on the inheriting contracts that just relay the call up the inheritance chain) but that's worse than just using another interface imo.


src/fund/hub/Spoke.sol
```diff
@@ -41,6 +45,15 @@ contract Spoke is DSAuth {
         setOwner(address(0));
     }
 
-    function priceSource() public view returns (address) { return hub.priceSource(); }
+    function priceSource() public view override returns (address) { return hub.priceSource(); }
     function fundFactory() public view returns (address) { return routes.fundFactory; }
+    function getHub() public view override returns (IHub) { return IHub(address(hub)); }
+    function getRoutes()
```
SeanJCasey: Is this necessary because we can't override a public struct storage var?

src/fund/shares/ISharesToken.sol
```diff
@@ -1,34 +1,5 @@
...
-    function initialized() external view returns (bool);
-    function routes() external view returns (IHub.Routes memory);
-
-    // FUNCTIONS
-    function engine() external view returns (address);
-    function mlnToken() external view returns (address);
-    function priceSource() external view returns (address);
-    function fundFactory() external view returns (address);
+interface ISharesToken {
+    function createFor(address, uint256) external;
```
SeanJCasey: Just making a note that if `destroyFor` is never used externally, that function can/should be removed. Can do in this PR or another one.

So maybe this is a good example of where this interface paradigm makes understanding the state-changing access control surface area easier to monitor

src/fund/hub/Spoke.sol
travs: Exactly, though I think this will be [fixed in an upcoming version of solidity](https://github.com/ethereum/solidity/issues/8489)

src/fund/shares/ISharesToken.sol
travs: Yup I can do it in this one

PR #979, opened 2020-04-09
Add userdoc and devdoc output.

fubhy:


travs:
We will have this when we change to truffle to compile

PR #985, opened 2020-04-29
feat: chai integration

SeanJCasey:


fubhy:
@SeanJCasey I am working on merging this one too. Quite a few conflicts. Any comments before I dive in further?

EDIT: Nevermind. Too many contract changes that I don't grasp yet. This one is for you @SeanJCasey

fubhy:
@SeanJCasey 

- [x] Mock integratee
- [x] Mock price source
- [x] Deployment updated to feature both
- [ ] Testing

tests/contracts/integratees/MockChaiIntegratee.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity 0.6.8;
+
+import "../PreminedToken.sol";
+import "./utils/MockIntegrateeBase.sol";
+
+contract MockChaiIntegratee is MockIntegrateeBase, PreminedToken("Chai", "CHAI", 18) {
+    address public dai;
```
SeanJCasey: It doesn't matter since it's a test, but this could be `immutable DAI`

tests/contracts/integratees/MockChaiIntegratee.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity 0.6.8;
+
+import "../PreminedToken.sol";
+import "./utils/MockIntegrateeBase.sol";
+
+contract MockChaiIntegratee is MockIntegrateeBase, PreminedToken("Chai", "CHAI", 18) {
```
SeanJCasey: Interesting idea to just allow the defi protocol integratee to be an ERC20 token. This will work here for Chai since it's the only asset, but it wouldn't work for Compound or most other integrations b/c multiple tokens.

For the other protocols, we'll probably we'll need to keep a list of derivatives in the deployment script and seed the integratees with those just like primitives.

If that's the case, then it might be good to start that practice now so it's easy for the next devs working on derivative tests to use the same implementation.

What do you think?

tests/contracts/integratees/utils/MockIntegrateeBase.sol
```diff
@@ -69,11 +70,48 @@ abstract contract MockIntegrateeBase is NormalizedRateProviderBase {
...
+                _trader.transfer(amount);
             }
             else {
-                ERC20(asset).transfer(msg.sender, amount);
+                ERC20(asset).transfer(_trader, amount);
             }   
         }
     }
+
+    function __getRateAndSwapAssets(
```
SeanJCasey: I see why you would want to lift this logic, but I don't think this should be in `MockIntegrateeBase`, because some integratees will have multiple incoming or outgoing assets (0x, Uniswap LPs, etc).

It's true that several can use this exact same function though. Maybe create a `MockSimpleIntegrateeBase`?

contracts/prices/derivatives/ChaiPriceSource.sol
```diff
@@ -0,0 +1,51 @@
...
+        uint256 chi = (now > pot.rho()) ? pot.drip() : pot.chi();
+        return chi / 10 ** 9; // Refactor of mul(chi, 10 ** 18) / 10 ** 27
+    }
+}
+
+/// @notice Limited interface for Maker DSR's Pot contract
+/// @dev See DSR integration guide: https://github.com/makerdao/developerguides/blob/master/dai/dsr-integration-guide/dsr-integration-guide-01.md
+interface IPot {
+    function chi() external returns (uint256);
+    function rho() external returns (uint256);
```
SeanJCasey: I'm just realizing that `chi()` and `rho()` are both public vars with getters built in, so can you please add `view` to these in the interface?

tests/contracts/priceSources/MockChaiPriceSource.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity 0.6.8;
+
+contract MockChaiPriceSource {
+    uint256 public chi;
+    uint256 public rho;
+    uint256 public drip;
```
SeanJCasey: I'd be interested to see if this works, because because `drip()` is a non-view function that updates state.

I think the best thing to do is to get rid of this and `setDrip()`, and define a `function drip() external` that does nothing. (or if you're feeling ambitious and want to actually have it update `chi` and `rho`, you could do that also! Not sure if that's necessary for unit tests though?)

Here's the actual `drip()` function: https://github.com/makerdao/dss/blob/b280b98e0c661d5d8effe2e9626e51978dab33ce/src/pot.sol#L140

And this is the `Pot` contract documentation in case that's helpful: https://docs.makerdao.com/smart-contract-modules/rates-module/pot-detailed-documentation

fubhy: If you can give this another review before you start your deep dive in the upgradability stuff that would be great. Once you you give the green light for the mock contracts I'll start with the tests.

fubhy:
@SeanJCasey We can't actually simplify this because the ChaiAdapter requires Chai and the Chai Integratee to be identical (because of how it handles allowance, transfer handler validation, etc.). Hence I switched back and simplified the MockIntegratee to actually simply be a PreminedToken and use burn/mint.

fubhy:
@SeanJCasey Furthermore, I believe the allowance() logic in the ChaiAdapter.sol was previously incorrect. Can you double check that? Imho the Adapter needs to have allowance set for CHAI, not the msg.sender because that's handled via the Vault (Vault => Adapter => Exchange and back) in the transfer handler helper, no?

Of course we could simplify this as already described in the TODO statement for exit(). But it then would have to be simplified for both. Until then, I think my change is correct.

fubhy (approved):

PR #986, opened 2020-05-06
feat: abstract PolicyManager and policies to allow extensible and sharable policies

SeanJCasey:
This PR makes the PolicyManager more versatile and extensible, specifically in order to pave the way to handle various kinds of vault integration adapters. It follows a similar structure to the abstracted `Vault().callOnIntegrations()`, in that it simply takes a `bytes _encodedArgs` param, the encoding schema of which is particular to each policy. It also implements a hook system by which a policy specifies where (`PolicyHook`) and when (`PolicyHookExecutionTime`) it should be executed.

This PR also makes policies sharable between funds, so that the official policies are registered to the Registry, and are then simply enabled on the fund, with fund-specific config stored on the policy itself.

There are also significant changes to how risk management arguments are provided in `callOnIntegration`:
- lift checking if an asset is registered to `callOnIntegration` (from individual adapters)
- lift policy validation calls to `callOnIntegration` (from individual adapters)
- have each adapter provide a list of expected incoming assets for a transaction, which is callable from `callOnIntegration`
- calculate all incoming and outgoing assets in a `callOnIntegration` by comparing pre-/post-action balances, and use these values for risk management policies
- refactor policies to use incoming / outgoing assets and amounts, rather than specific things like 'makerAsset' and 'takerAsset'


SeanJCasey:
In each policy, we currently don't have an external function that allows checking a rule against a specified fund with specified args. If we wanted to, we could include a `passesRule()` function that allows testing of the rules for a given fund with certain parameters, and then `validateRule()` (for a fund) would simply pass the `msg.sender` and args to `passesRule()`. It's not strictly necessary, so I didn't include it for now.

src/fund/policies/PolicyManager.sol
```diff
@@ -10,95 +11,108 @@ import "./IPolicyManager.sol";
...
+            "enablePolicies: array lengths unequal"
+        );
+
+        // Enable each policy with settings
+        IRegistry registry = __getRegistry();
+        for (uint256 i = 0; i < _policies.length; i++) {
+            IPolicy policy = IPolicy(_policies[i]);
+            require(
+                registry.policyIsRegistered(address(policy)),
+                "enablePolicies: Policy is not on Registry"
```
travs, line 47: Might be helpful to have the address of the failing policy in the revert message like we do in other contexts. Also for the require directly below (`__policyIsEnabled`)

src/fund/policies/utils/AddressListPolicyMixin.sol
```diff
@@ -0,0 +1,67 @@
...
+    // INTERNAL FUNCTIONS
+
+    /// @notice Helper to add addresses to the calling fund's list
+    function __addToList(address[] memory _items) internal {
+        require(_items.length > 0, "__addToList: no addresses provided");
+
+        for (uint256 i = 0; i < _items.length; i++) {
+            require(
+                EnumerableSet.add(policyManagerToList[msg.sender], _items[i]),
+                "__addToList: address already exists in list"
```
travs, line 47: maybe adding the address to the error message here would help for debugging as well

src/fund/policies/utils/AddressListPolicyMixin.sol
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /// @notice Helper to remmove addresses from the calling fund's list
+    function __removeFromList(address[] memory _items) internal {
+        require(_items.length > 0, "__removeFromList: no addresses provided");
+
+        for (uint256 i = 0; i < _items.length; i++) {
+            require(
+                EnumerableSet.remove(policyManagerToList[msg.sender], _items[i]),
+                "__removeFromList: address does not exist in list"
```
travs, line 61: and here

src/fund/policies/risk-management/MaxPositions.sol
```diff
@@ -1,35 +1,48 @@
...
-    function rule(bytes4 sig, address[5] calldata addresses, uint[3] calldata values, bytes32 identifier)
+    // EXTERNAL FUNCTIONS
+
+    /// @notice Add the initial policy settings for a fund
+    /// @param _encodedSettings Encoded settings to apply to a fund
+    /// @dev A fund's PolicyManager is always the sender
+    /// @dev Only called once, on PolicyManager.enablePolicies()
+    function addFundSettings(bytes calldata _encodedSettings) external override onlyPolicyManager {
+        uint256 maxPositions = abi.decode(_encodedSettings, (uint256));
+        require(maxPositions > 1, "addFundSettings: maxPositions must be greater than 1");
```
travs, line 25: What if the manager just wants one position? Not sure if there's a sensible use-case for that but should they have the option?

src/fund/policies/risk-management/PriceTolerance.sol
```diff
@@ -1,121 +1,106 @@
...
-    {
-        IPriceSource pricefeed = IPriceSource(
-            IRegistry(IHub(Spoke(msg.sender).HUB()).REGISTRY()).priceSource()
+    /// @notice Add the initial policy settings for a fund
+    /// @param _encodedSettings Encoded settings to apply to a fund
+    /// @dev A fund's PolicyManager is always the sender
+    /// @dev Only called once, on PolicyManager.enablePolicies()
+    function addFundSettings(bytes calldata _encodedSettings) external override onlyPolicyManager {
+        uint256 priceTolerance = abi.decode(_encodedSettings, (uint256));
+        require(priceTolerance > 0, "addFundSettings: priceTolerance must be greater than 0");
```
travs: Maybe it's very strict, but isn't `0` valid here?

travs (changes requested):

src/fund/policies/PolicyManager.sol
SeanJCasey: We'd need to implement a library for concatenating string + address. Are we doing this somewhere else?

src/fund/policies/utils/AddressListPolicyMixin.sol
SeanJCasey: Same comment as above (wouldn't we need a lib for that concatenation?)

src/fund/policies/utils/AddressListPolicyMixin.sol
SeanJCasey: Same comment again :P

src/fund/policies/risk-management/MaxPositions.sol
SeanJCasey: I don't think so, because the denomination asset is included, so the fund would be unusable (investors send asset1, and the fund manager can't make any trades/loans/etc for an asset2)

src/fund/policies/PolicyManager.sol
travs: Ah yeah, I was looking at this type of thing, also in PolicyManager.sol.

```
                    string(abi.encodePacked(
                        "Rule evaluated to false: ",
                        IPolicy(policies[i]).identifier()
                    ))
```

I guess it doesn't work with addresses though

src/fund/policies/risk-management/PriceTolerance.sol
SeanJCasey: Tough, because then we can't validate the input, but after discussing with the others, I think you're right about allowing `0` as a valid amount and then just leaving it up to frontends to properly validate their inputs.

SeanJCasey: Fixed and pushed

travs (approved):

PR #987, opened 2020-05-07
Fix: support tokens with incomplete ERC20 implementation

luongnt95:


tests/integration/fundBadERC20Redemption.test.js
```diff
@@ -0,0 +1,83 @@
...
+      investor,
+      tokenContract: omg,
+    },
+    manager,
+    quoteToken: weth.options.address,
+    fundFactory
+  });
+});
+
+describe('Investor invests BadERC20 token to a fund', () => {
```
SeanJCasey: Probably don't need a `describe` block, since it's only 1 test block below. You can remove this, or edit to "investor _redeems_ BadERC20 token _from_ a fund"

src/dependencies/token/IBadERC20.sol
```diff
@@ -0,0 +1,7 @@
+pragma solidity 0.6.4;
+
+interface IBadERC20 {
```
SeanJCasey: Can we name this something like `FlexibleIERC20` and add a description in Natspec to the top about what it does? I just don't like having the `Bad~` namespacing with our prod contracts lol

SeanJCasey: The `BadERC20Token.sol` naming is fine, because that accurately describes what it is and it's only in tests, but our interface can handle both that and normal ERC20 tokens.

tests/integration/fundBadERC20Redemption.test.js
```diff
@@ -0,0 +1,83 @@
...
+    const { shares, vault } = fund;
+
+    const fundBadERC20Balance = new BN(await call(omg, 'balanceOf', [vault.options.address]));
+    expect(fundBadERC20Balance).bigNumberEq(new BN(toWei('1', 'ether')));
+
+    const preInvestorShares = new BN(await call(shares, 'balanceOf', [investor]));
+    expect(preInvestorShares).bigNumberGt(new BN(0));
+
+    await send(shares, 'redeemShares', [], investorTxOpts);
+
```
SeanJCasey, line 71: Let's assert the user OMG balance also (postInvestorBadERC20Balance - preInvestorBadERC20Balance). It's almost def working, but good to be 100% sure.

SeanJCasey (changes requested):
Nice job @luongnt95 ! Just a couple minor naming edits and an additional test assertion.

src/dependencies/token/IERC20Flexible.sol
```diff
@@ -0,0 +1,11 @@
+pragma solidity 0.6.4;
+
+/// @title Flexible ERC20 interface
+/// @author Melon Council DAO <security@meloncoucil.io>
+/// @notice support standard ERC20 and BadERC20
+/// @notice A 'BadERC20Token' token is one that uses an old version of the ERC20 standard. Basically, this version does not return anything from `transfer` and `transferFrom`, whereas most modern implementions of ERC20 return a boolean to indicate success or failure.
```
SeanJCasey: @luongnt95 We realized recently that we should not have multiple of the same tag (e.g., `@dev` or `@notice`). So this could be:

```
/// @notice Here is the first comment.
/// Here is the second comment.

SeanJCasey (changes requested):
Just one tiny, tiny change

SeanJCasey (approved):

PR #988, opened 2020-05-13
Fix: remove partial redemption

luongnt95:


src/fund/shares/Shares.sol
```diff
@@ -148,7 +148,7 @@ contract Shares is IShares, TokenUser, Spoke, SharesToken {
         // assigned for fund manager (and potentially other fee recipients in the future)
         IFeeManager(routes.feeManager).rewardAllFees();
 
-        redeemSharesQuantity(balanceOf(msg.sender));
+        redeemSharesQuantity(balanceOf(msg.sender), true);
```
SeanJCasey: This should be `false`. If the transaction fails, then they can use `redeemSharesQuantity` with `true`

tests/integration/fundMaliciousRedemption.test.js
```diff
@@ -1,9 +1,7 @@
 /*
  * @file Tests fund's ability to handle a malicious redemption attempts
  *
- * @test Redeem fails when malicious token is present
- * @test Redemption with non-registered assets does not succeed
- * @test redeemSharesWithConstraints succeeds to withdraw specific assets only
+ * @test Redeem succeeds when malicious token is present
```
SeanJCasey: Let's organize this as 3 `test` blocks:

1. "Fund receives Malicious token" (you can keep as an investment for now, but this will need to change when only denomination asset investments are allowed. If you want to change this to a fund taking a trade, you can do that too.) -- fund receives token, then token starts reverting
2. "redeemShares fails"
3. "redeemSharesQuantity succeeds when _bypassFailure is set to true"

src/fund/shares/Shares.sol
```diff
@@ -251,68 +251,54 @@ contract Shares is IShares, TokenUser, Spoke, SharesToken {
     /// @notice Redeem a specified quantity of the sender's shares
     /// for a proportionate slice of the fund's assets
     /// @param _sharesQuantity The amount of shares to redeem
```
SeanJCasey, line 271: 1. Add `/// @param _bypassFailure True if token transfer failures should be ignored and forfeited`
2. Add `/// @dev If _bypassFailure is set to true, the user will lose their claim to any assets for which the transfer function fails. This should always be false unless explicitly intended`

src/fund/shares/Shares.sol
```diff
@@ -251,68 +251,54 @@ contract Shares is IShares, TokenUser, Spoke, SharesToken {
...
 
-            // Confirm asset has not already been transfered
-            for (uint256 j = 0; j < i; j++) {
+            // Transfer owed asset to redeemer
+            try vault.withdraw(ownedAssets[i], ownedQuantities[i]) {}
+            catch {}
+
+            uint256 receiverPreBalance = IERC20(ownedAssets[i]).balanceOf(msg.sender);
+            try IERC20Flexible(ownedAssets[i]).transfer(msg.sender, ownedQuantities[i]) {
+                uint256 receiverPostBalance = IERC20(ownedAssets[i]).balanceOf(msg.sender);
```
SeanJCasey: I don't think we need to validate the pre/post-transfer balances here, because there is nothing we can do about a fund asset that does not actually transfer tokens like it is supposed to. We only need to protect against an asset that is unredeemable.

SeanJCasey (changes requested):
Nice job @luongnt95 , I have just a couple changes in the code, adding some natspec comments, and adding to the test suite so that we test both failure and success conditions.

SeanJCasey:
@luongnt95 Actually, I think we should structure these functions slightly differently, to protect against bad / accidental inputs.

1. `__redeemShares(uint256 _sharesQuantity, bool _bypassFailure) private` - basically the current `redeemSharesQuantity()`
2. `redeemShares() external` - calls `__redeemShares(totalShares, false)`
3. `redeemSharesQuantity(uint256 _sharesQuantity) external` - calls `__redeemShares(_sharesQuantity, false)`
4. `redeemSharesEmergency() external` - calls `__redeemShares(totalShares, false)`

Most of the comments in the previous review are still valid (I will close the ones that are not).

src/fund/shares/Shares.sol
SeanJCasey: Ok, so since we're changing to the new functions, we don't need the `@param`, but `redeemSharesEmergency` should include the `@dev` comment

src/fund/shares/Shares.sol
SeanJCasey: This is no longer relevant since we're changing the functions

src/fund/shares/Shares.sol
```diff
@@ -248,76 +260,58 @@ contract Shares is IShares, TokenUser, Spoke, SharesToken {
...
-            // Confirm asset has not already been transfered
-            for (uint256 j = 0; j < i; j++) {
-                require(
-                    _assets[i] != _assets[j],
-                    "Attempted to redeem duplicate asset"
-                );
-            }
-
+        // Calculate and transfer owned assets to redeemer
+        uint256[] memory ownedQuantities = new uint256[](ownedAssets.length);
```
SeanJCasey: This actually isn't the "owned" quantity (the amount the fund owns), it's the "owed" quantity (the amount the user gets paid out). The namespacing is probably too similar between "owned" and "owed" though. Let's change this to `payoutAssets` and `payoutQuantities`?

tests/integration/fundMaliciousRedemption.test.js
```diff
@@ -147,13 +91,20 @@ test('General redeem fails in presence of malicious token', async () => {
...
   // Activate malicious token
   await send(maliciousToken, 'startReverting', [], defaultTxOpts);
-  
+});
+
+test('redeemShares fails in presence of malicious token', async () => {
+  const { shares } = fund;
+
+  // Activate malicious token
+  await send(maliciousToken, 'startReverting', [], defaultTxOpts);
```
SeanJCasey: There are 2 `startReverting` calls in this file

src/fund/shares/Shares.sol
```diff
@@ -147,8 +147,20 @@ contract Shares is IShares, TokenUser, Spoke, SharesToken {
...
         // assigned for fund manager (and potentially other fee recipients in the future)
         IFeeManager(routes.feeManager).rewardAllFees();
+        __redeemShares(balanceOf(msg.sender), false);
+    }
+
+    /// @notice Redeem all of the sender's shares for a proportionate slice of the fund's assets
+    /// @dev _bypassFailure is set to true, the user will lose their claim to any assets for
+    /// which the transfer function fails.
+    function redeemSharesEmergency() external {
+        __redeemShares(balanceOf(msg.sender), true);
```
SeanJCasey, line 158: Add `IFeeManager(routes.feeManager).rewardAllFees();` before `__redeemShares()`

src/fund/shares/Shares.sol
```diff
@@ -147,8 +147,20 @@ contract Shares is IShares, TokenUser, Spoke, SharesToken {
...
+    }
 
-        redeemSharesQuantity(balanceOf(msg.sender));
+    /// @notice Redeem a specified quantity of the sender's shares
+    /// for a proportionate slice of the fund's assets
+    function redeemSharesQuantity(uint256 _sharesQuantity) external {
```
SeanJCasey, line 164: Add `@param` Natspec

src/fund/shares/Shares.sol
SeanJCasey: (remember to change "owned" in the comment on line 288, too)

SeanJCasey (changes requested):
Looking good @luongnt95 , just a few minor changes.

SeanJCasey (approved):
Nice job! Will merge now.

PR #989, opened 2020-05-18
refactor: change how FundFactory and a fund's Hub setup, store, and use fund addresses

SeanJCasey:
This PR primarily refactors how fund addresses are stored and accessed in a Fund, and how they are set during fund creation by the FundFactory.

- update the Hub and Spoke contracts to primarily serve as routers, with convenience getters and modifiers for other contracts
- completely remove `DSAuth.sol` and `DSGuard.sol` permissioning from fund contracts and replace with modifiers
- unify a fund's status into a `FundStatus` enum field
- update `FundFactory` pipeline to add each component address to its `Hub` immediately upon availability
- apply coding and commenting standards to `FundFactory` 
- update `FundFactory` namespaces to be specific to fund creation, to prepare for possibly merging it with `Registry`


src/engine/Engine.sol
```diff
@@ -173,21 +174,16 @@ contract Engine is IEngine, DSMath {
...
-            // 1. Spoke points to hub
-            // 2. Hub confirms it is a spoke
-            // 3. Fund exists for hub
-            try ISpoke(_who).getHub() returns (IHub hub) {
-                return hub.isSpoke(_who) && registry.fundIsRegistered(address(hub));
-            }
-            catch {
-                return false;
-            }
+    /// @notice Helper to check whether an address is a valid fund Vault
```
fubhy, line 177: Shouldn't that be @dev

src/fund/hub/Hub.sol
```diff
@@ -9,72 +8,95 @@ import "./IHub.sol";
...
-    bool public override fundInitialized;
-    uint public creationTime;
-    mapping (address => bool) public override isSpoke;
+    event SharesSet(address);
 
-    modifier onlyCreator() {
-        require(msg.sender == creator, "Only creator can do this");
+    event VaultSet(address);
+
+    event StatusUpdated(FundStatus);
```
fubhy: Would it make sense to have dedicated events for each update? What avout indexed event parameters? When do we do that?

src/fund/hub/Hub.sol
```diff
@@ -9,72 +8,95 @@ import "./IHub.sol";
...
+
+    // Fund vars
+    address public override MANAGER;
+    string public NAME;
+    FundStatus public override status;
+
```
fubhy, line 26: why are these sometimes lower case and sometimes upper case?

src/fund/hub/Hub.sol
```diff
@@ -9,72 +8,95 @@ import "./IHub.sol";
...
+        MANAGER = _manager;
+        NAME = _name;
+        REGISTRY = _registry;
     }
 
-    function shutDownFund() external {
-        require(msg.sender == manager, "shutDownFund: Only fund manager can call this function");
-        isShutDown = true;
-        emit FundShutDown();
+    /// @notice Initializes a fund (actives it)
```
fubhy: Activates typo

src/fund/hub/Hub.sol
```diff
@@ -9,72 +8,95 @@ import "./IHub.sol";
...
+    function initializeFund() external onlyFundFactory {
+        require(status == FundStatus.Draft, "initialize: Fund already initialized");
```
fubhy: That's a weird error message

src/fund/hub/Hub.sol
```diff
@@ -9,72 +8,95 @@ import "./IHub.sol";
...
-            routes.policyManager,
-            bytes4(keccak256('batchRegister(bytes4[],address[])'))
-        );
-        permit(manager, routes.policyManager, bytes4(keccak256('register(bytes4,address)')));
-        permit(routes.feeManager, routes.shares, bytes4(keccak256('createFor(address,uint256)')));
-        fundInitialized = true;
+    /// @notice Sets the feeManager address for the fund
+    /// @param _feeManager The FeeManager component for the fund
+    function setFeeManager(address _feeManager) external onlyFundFactory {
+        require(feeManager == address(0), "setFeeManager: Already set");
```
fubhy: Also weird error message

fubhy: also all followong ones

src/engine/Engine.sol
```diff
@@ -173,21 +174,16 @@ contract Engine is IEngine, DSMath {
...
+    function __isValidVault(address _who) private view returns (bool) {
+        // 1. Spoke points to hub
+        // 2. Hub is valid
+        // 3. Hub points to vault
+        try Spoke(_who).HUB() returns (address hubAddress) {
```
fubhy: Why not ISpoke here

src/fund/fees/PerformanceFee.sol
```diff
@@ -33,7 +33,7 @@ contract PerformanceFee is DSMath {
 
     /// @notice Assumes management fee is zero
     function feeAmount() external returns (uint feeInShares) {
-        Shares shares = Shares(IHub(Spoke(msg.sender).getHub()).shares());
+        Shares shares = Shares(IHub(Spoke(msg.sender).HUB()).shares());
```
fubhy, line 36: Isnt HUB on the spoke interface?

src/fund/policies/risk-management/PriceTolerance.sol
```diff
@@ -41,7 +41,9 @@ contract PriceTolerance is TradingSignatures, DSMath {
         view
         returns (bool)
     {
-        IPriceSource pricefeed = IPriceSource(Spoke(msg.sender).priceSource());
+        IPriceSource pricefeed = IPriceSource(
```
fubhy, line 44: Helper getter functions for policies?

fubhy (dismissed):

src/engine/Engine.sol
SeanJCasey: @fubhy Natspec isn't parsed for internal/private functions, so we had decided just to use `@notice` tags for general descriptions of these functions. You're right that it might be more correct to use `@dev`, I'd just want to confirm that newlines without a decorator are parsed as the previous decorator (rather than as `@notice` by default, which is what the docs say happens to undecorated lines).

I'm thinking we can leave these as `@notice` for consistency with our codebase for now, but we can easily change all these in a future PR.

src/fund/hub/Hub.sol
SeanJCasey: We can index the event return value, `event StatusUpdated(FundStatus indexed status);`

I think this solves the problem, or do we want/need separate events?

src/fund/hub/Hub.sol
SeanJCasey: As discussed in our chat, UPPER_CASE is used for constants or vars set in the constructor. The latter are storage vars at the moment, but will be set to `immutable` (and exist in deployed bytecode) when we upgrade to a solidity release that supports it.

src/fund/hub/Hub.sol
SeanJCasey: Fixed

src/fund/hub/Hub.sol
SeanJCasey: This is how we're formatting error messages everywhere in the protocol for now:

`[functionName]: [description]`

It's helpful for us to easily identify which function contains offending code, and we use it consistently, so I'll mark this as resolved until we start the larger task of implementing error status codes.

src/fund/hub/Hub.sol
SeanJCasey: Addressed this in another comment

src/engine/Engine.sol
SeanJCasey: Updated to `ISpoke` (and `IHub`)

src/fund/fees/PerformanceFee.sol
SeanJCasey: Because of how we're using interfaces (as representations of the surface area of interactions between core contracts), we don't want to add new functions to an interface based on the needs of plugin contracts. It would be fine in this case (for this function), but in principle we're using full contract imports in plugins so that we don't accidentally for a function into the interface that doesn't need to be there.

src/fund/policies/risk-management/PriceTolerance.sol
SeanJCasey: @fubhy Came up with a nice solution here: https://github.com/melonproject/protocol/pull/989/commits/ec508e584ad6e06a33bf9d825171c1c948f502dd

I abstracted routing into a `FundRouterMixin`, which I think is important because it guarantees that routing logic happens in a fund and in plugins in exactly the same way.

We can either use this directly in plugins and pass `ISpoke(msg.sender).HUB()` in as an argument (e.g., `IPriceSource(__getPriceSource(ISpoke(msg.sender).HUB()))`), or create a `SpokeCallee` contract that inherits `FundRouterMixin` and will simplify this to `IPriceSource(__getPriceSource())`, `IRegistry(__getRegistry())`, etc. I think that latter option will be the way to go.

I will implement this in #986 (no sense in doing it to policies and fees now, as this will all change shortly).

src/fund/hub/Hub.sol
luongnt95: @SeanJCasey @fubhy  I think it's better to have a full and meaningful message error like: "setFeeManger: feeManager is already set". It's clearer for newbies

src/fund/hub/FundRouterMixin.sol
```diff
@@ -0,0 +1,33 @@
+pragma solidity 0.6.4;
+
+import "../../registry/IRegistry.sol";
+import "./IHub.sol";
+
+/// @title FundRouterMixin Contract
+/// @author Melon Council DAO <security@meloncoucil.io>
+/// @notice A mixin that provides routing for a given fund
+abstract contract FundRouterMixin {
```
luongnt95, line 9: why not make these functions `public` instead so that both plugin and component contracts can use.

src/fund/policies/risk-management/PriceTolerance.sol
luongnt95: I think we just need to create getter functions in hub public so that plugin can call these functions via hub. It's quite redundant to have a mixin here because it will add more code to the plugin contract

luongnt95 (changes requested):

src/fund/hub/FundRouterMixin.sol
```diff
@@ -0,0 +1,33 @@
...
+    function __getFeeManager(address _hub) internal view returns (address) {
```
luongnt95, line 10: why not make these functions public instead so that both plugin and component contracts can use.

luongnt95 (changes requested):

src/fund/hub/FundRouterMixin.sol
SeanJCasey: @luongnt95 We could, but `public` functions take up much more space in bytecode than `internal` functions, because it creates interfaces to interact with the contract. Especially for fund contracts (which we deploy over and over again, per fund), we want to optimize space by only providing necessary external functions. (`public` is actually even worse than external in terms of efficiency, because it uses memory instead of calldata).

So in this case, we only provide the essential internal functions for consistent routing, which can then be implemented by other inheriting contracts (this is what policies and fees will do).

Remember that routes are already available via the hub, so these are purely convenience functions.

src/fund/hub/Hub.sol
SeanJCasey: @luongnt95 I updated the revert strings to your suggestion.

luongnt95 (approved):

PR #990, opened 2020-05-21
feat: add a ValueInterpreter infrastructural contract to provide value conversions between assets

SeanJCasey:
Currently, price-related conversions happen sporadically across the protocol, and rely heavily on code within the KyberPriceFeed itself.

This PR abstracts all interpreting of prices and how to apply conversion rates into an infrastructural `ValueInterpreter` contract. It also lays the foundation for converting from an asset to a derivative asset (e.g., cDAI) and vice-versa by looking only at price source interfaces, rather than at specific price sources. This contract is part of the infrastructural layer, and its deployed address is stored in the Registry.


src/prices/KyberPriceFeed.sol
```diff
@@ -125,148 +125,67 @@ contract KyberPriceFeed is IPriceSource, DSMath {
...
-    /// @notice Return getPrice for each of _assets
-    /// @param _assets Assets for which prices should be returned
-    /// @return prices_ Prices for each of the assets_
-    /// @return timestamps_ Update times for each of the assets_
-    function getPrices(address[] calldata _assets)
+    /// @notice Returns rate and validity for some pair of assets using pricefeed prices
+    /// @param _baseAsset Address of base asset from the pair
+    /// @param _quoteAsset Address of quote asset from the pair
+    /// @return rate_ The price of _baseAsset in terms of _quoteAsset
+    /// @return isValid_ True if the rate for this pair is passes validation checks
```
travs: small sentence thing (is passes)

SeanJCasey:
@travs I just pushed a commit that refactored some stuff, most importantly handling multiple underlying assets and recursion (when an underlying asset is itself a derivative). The naming is getting potentially confusing, so wanted to leave you a few notes on terminology:

"Assets," "primitives," and "derivatives":
- I'm using "asset" to refer to any and all tokens in our system
- I'm using "primitive" (in one function name only) to refer to all tokens that are in the Registry, which we store prices for in the primary pricefeed (e.g., WETH, DAI, MLN, etc)
- I'm using "derivative" to refer to all tokens that are registered as "derivatives" and represent a claim to underlying primitives (e.g., cDAI -> DAI)

"Underlying":
- These are the tokens that a derivative is redeemable for. I'm calling them "underlying" rather than "primitives", because 1. that's what Compound calls them, but more importantly 2. we don't immediately know if an underlying token is a primitive or another derivative.

You'll see this terminology in `IDerivativePriceSource.getRatesForUnderlyings()`, which returns the multiple rates to convert an amount of a derivative into the multiple underlying tokens.

src/prices/ValueInterpreter.sol
```diff
@@ -0,0 +1,155 @@
...
+        ) = IDerivativePriceSource(derivativePriceSource).getRatesToUnderlyings(_derivative);
+
+        for (uint256 i = 0; i < underlyings.length; i++) {
+            uint256 underlyingAmount = __calcConversionAmount(underlyings[i], _amount, rates[i]);
+            (
+                uint256 underlyingValue,
+                bool underlyingIsValid
+            ) = __calcAssetValue(underlyings[i], underlyingAmount, _quoteAsset, _useLiveRate);
+
+            if (isValid_ && !underlyingIsValid) isValid_ = false;
```
travs: can the conditional be shortened to just `if (!underlyingIsValid)`? Also, is it possible for `isValid_` to be returned as `true`, even with the change I mentioned?

travs (changes requested):
Looks good to me, just that one return value I commented on.

src/prices/KyberPriceFeed.sol
SeanJCasey: Fixed below

src/prices/ValueInterpreter.sol
SeanJCasey: Ah yes, I forgot to assign `isValid_` to `true` before the loop. With L142, I was trying to not re-assign to `false` redundantly if there are multiple invalid rates, but it's clearer with your suggestion, so I've done that.

Fixed below.

SeanJCasey:
@travs A few final changes / addendums here:

- Refactor "asset" nomenclature to "primitive" in the `Registry`
- Create subdirectory for `prices/primitives/`
- Remove `view` from `IDerivativePriceSource.getRatesToUnderlyings()`, since some of these calculations could update state on a 3rd party protocol (e.g., DSR)

travs (approved):

PR #991, opened 2020-05-27
fix: assert that ERC20 balances change correctly when redeeming shares

luongnt95:


SeanJCasey (approved):
Nice, thanks @luongnt95 !

PR #992, opened 2020-05-29
feat: abstract fee management and restrict shares creation to Shares

SeanJCasey:
This PR is basically a complete re-write of `FeeManager` and the way that fees work, including re-writes of `ManagementFee` and `PerformanceFee` as well. The goal is abstraction, where the FeeManager is not hardcoded with idiosyncrasies to suit any fee (or combination of particular fees), but rather it can facilitate a wide range of potential fee implementations.

Rather than imposing knowledge of particular fees on the `FeeManager` (e.g., fee #1 is a management fee and fee #2 is a performance fee, and must be called in this order), the `FeeManager` is un-opinionated about the identities of particular fees. Instead, it only cares about the fee's type (its `FeeHook`), which is specified by the caller of fee settlement (e.g., `Shares` will call `FeeManager.settleFees()` with `FeeHook.BuyShares` when buying shares and with `FeeHook.Continuous` before any change to `Shares.totalSupply` (i.e., any attempt to buy or sell shares).

The main point of complexity is the way that fees are distributed, allowing for shares due to be optionally held for payout at a later time. For policies such as `PerformanceFee`, we want to continuously add/subtract from a running count of "shares outstanding", so that frequent entries/exits of investors do not create an unfair handicap for late investors (i.e., paying for performance they did not receive). The `FeeManager` thus gives fees the chance to respond to two calls: "settlement" and "payout of shares outstanding". Some fees such as `ManagementFee` only implement settlement and distribute shares to the fund manager (or any recipient) immediately, which others such as `PerformanceFee` take shares into the fee's custody until they are able to be unlocked by some condition. The `FeeManager` serves as the coordinator of these distributions, differentially minting and burning shares to/from the various parties as dictated by the fees (again, the FeeManager stays un-opinionated about the parties and amount involved).

The individual fees themselves have been re-structured into base contracts that provide helper functions and setup common to a particular type of fee (only `ContinuousFeeBase` for this PR, but will also need one for `BuySharesFeeBase`).

It should be noted that `PerformanceFee` is likely to change heavily, and should be considered a work-in-progress at this stage.

Also, there are no policy implementations for `FeeHook.BuyShares`, which we will definitely want before the release.

### DIFF
--- a/contracts/factory/FundFactory.sol
+++ b/contracts/factory/FundFactory.sol
@@ -48,8 +48,7 @@ contract FundFactory is AmguConsumer {
         address[] adapters;
         address denominationAsset;
         address[] fees;
-        uint256[] feeRates;
-        uint256[] feePeriods;
+        bytes[] feeSettings;
         address[] policies;
         bytes[] policySettings;
     }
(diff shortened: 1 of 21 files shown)


fubhy:
@SeanJCasey I merged this with develop and fixed the conflicts.

fubhy (approved):

PR #997, opened 2020-06-01
feat: restrict updating melon council accounts used for access control

SeanJCasey:
This PR replaces DSAuth in the `Registry` with a `MelonCouncilOwnable` contract, a slightly modified child of OpenZeppelin's `Ownable` contract.

- `MTC` and `MGM` are `immutable` vars on the new contract, defined in the `constructor()`, and are then unchangable
- The contract owner is initially its deployer, and ownership can be transferred, but only until `MTC` and `_owner` are the same.
- Relinquishing ownership is prohibited.
- This assures that once the MTC has ownership of the contract, they will forever be the owner.

### DIFF
--- a/migrations/14_post_deployment.js
+++ b/migrations/14_post_deployment.js
@@ -1,5 +1,4 @@
 const mainnetAddrs = require('../config');
-const Registry = artifacts.require('Registry');
 const OasisDex = artifacts.require('IOasisDex');
 const WETH = artifacts.require('WETH');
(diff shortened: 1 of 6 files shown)


SeanJCasey:
The tests will fail, because the contract dependency won't compile from `node_modules`, but it's not worth changing this until we have merged in the Truffle structure.

fubhy (approved):

luongnt95 (approved):

PR #999, opened 2020-06-05
refactor: FundFactory adds Fund to Registry with required unique name

SeanJCasey:
Just a simple PR I split off from #986 , as it is not related to the the main objective of that PR.

Enforces unique fund naming in the `Registry`, and allows any known `FundFactory` to register funds (e.g., in case a fund starts the fund deployment process with `FundFactory A` and then `FundFactory B` is deployed and becomes the most recent FundFactory).


travs (approved):
Look good!

PR #1000, opened 2020-06-13
refactor: improve shares requests

SeanJCasey:


src/fund/shares/Shares.sol
```diff
@@ -59,45 +52,48 @@ contract Shares is IShares, TokenUser, Spoke, SharesToken {
...
+        uint256 _minSharesQuantity
     )
         external
         override
         onlySharesRequestor
-        returns (uint256 costInInvestmentAsset_)
+        returns (uint256 sharesBought_)
     {
-        costInInvestmentAsset_ = getSharesCostInAsset(_sharesQuantity, _investmentAsset);
+        address denominationAsset = DENOMINATION_ASSET;
```
travs: this is fine, but do we really need to assign from the constant?

src/requests/SharesRequestor.sol
```diff
@@ -117,158 +110,155 @@ contract SharesRequestor is DSMath, TokenUser, AmguConsumer {
...
-        // Do the asset transfers in a separate step after altering state
-        __transferRequestSurplusAssets(
-            _requestOwner,
-            request.investmentAsset,
-            request.maxInvestmentAmount,
-            investmentAmountFilled,
-            request.incentiveFee
-        );
+        // Reward sender with incentive
+        msg.sender.transfer(request.incentiveFee);
```
travs, line 124: Maybe I'm missing some permissioning but this is susceptible to frontrunning. I guess we will fix that later on?

travs (changes requested):
Looks good, just those couple notes

src/requests/SharesRequestor.sol
SeanJCasey: As discussed in our chat, we'll leave it as-is as this is by design, but we will want to revisit that design before the next release.

src/fund/shares/Shares.sol
SeanJCasey: Fixed

travs (approved):

PR #1003, opened 2020-06-28
refactor: externalize integration adapters and combine asset validation into callOnIntegration

SeanJCasey:
This PR changes the way adapters work by having the `Vault` call them via a low-level `.call()` rather than `.delegatecall()`. This has the affect of adapter functions being called within their own contract context, rather than the context of the caller (i.e., the `Vault`). This allows the `Vault` to treat adapters in a less trusted manner, as adapters no longer have direct access to manipulate its storage, and the Vault specifies which and how much asset to approve the adapter to use in each action (though these amounts are currently specified by the adapter, so it can still ask for whatever assets it wants to use).

A few other tasks and changes were made during this refactor:
- The validation functionalities of `OrderFiller.sol` and `OrderTaker.sol` are now lifted to the Vault and take place alongside the asset balance calculations made for policy management.
- All adapters inherit an `AdapterBase` with common operations. Especially important is the `fundAssetsTransferHandler` modifier that handles the transfer of approved assets to the adapter, and from the adapter back to the `Vault` at the end of the call. 
- Adapters are refactored to implement function calls (i.e., `takeOrder`) directly once again, rather than the convoluted syntax of `OrderTaker`.
- A `MathHelpers.sol` util was added to help with finding a missing quantity in two relative fractions (i.e., in an order). 
- An `AddressArrayLib.sol` library was added for common operations performed on asset arrays, e.g., checking uniqueness of the array and checking if an item is contained by the array.
- "maker/taker" terminology was replaced by "incoming/outgoing" in adapters where makes/takes are not relevant (Kyber/Uniswap)
- All non-OasisDex-specific tests using OasisDex were refactored to use more relevant exchanges (Kyber and 0x)
- "Maker/taker" terminology was replaced with "incoming/outgoing" in tests where the latter is more relevant


SeanJCasey:
@fubhy I made the refactor that we discussed: passing the parsed assets to transfer as an encoded arg to `_adapter.call()`

contracts/fund/vault/Vault.sol
```diff
@@ -288,77 +233,212 @@ contract Vault is IVault, TokenUser, Spoke {
...
+            spendAssets_.length == spendAssetAmounts_.length,
+            "__preProcessCoI: spend assets arrays unequal"
+        );
+        require(
+            incomingAssets_.length == minIncomingAssetAmounts_.length,
+            "__preProcessCoI: incoming assets arrays unequal"
+        );
+        require(spendAssets_.isUniqueSet(), "__preProcessCoI: duplicate spend asset detected");
+        require(
+            incomingAssets_.isUniqueSet(),
```
fubhy, line 283: I love these array utils.

contracts/integrations/adapters/UniswapAdapter.sol
```diff
@@ -1,179 +1,198 @@
...
+            address outgoingAsset,
+            uint256 outgoingAssetAmount
+        ) = __decodeCallArgs(_encodedCallArgs);
+
+        // Validate args
+        require(minIncomingAssetAmount > 0, "takeOrder: minIncomingAssetAmount must be >0");
+        require(outgoingAssetAmount > 0, "takeOrder: outgoingAssetAmount must be >0");
+        require(incomingAsset != outgoingAsset, "takeOrder: incomingAsset and outgoingAsset cannot be the same");
+        require(incomingAsset != address(0), "takeOrder: incomingAsset cannot be empty");
+        require(outgoingAsset != address(0), "takeOrder: outgoingAsset cannot be empty");
```
fubhy, line 91: Should this maybe be moved to a helper function on AdapterBase?

contracts/integrations/adapters/UniswapAdapter.sol
```diff
@@ -1,179 +1,198 @@
...
-            add(block.timestamp, 1)
-        );
+        address tokenExchange = IUniswapFactory(EXCHANGE).getExchange(_incomingAsset);
+        IUniswapExchange(tokenExchange)
+            .ethToTokenSwapInput
+            {value: _outgoingAssetAmount}
+            (
+                _minIncomingAssetAmount,
+                add(block.timestamp, 1)
+            );
```
fubhy, line 150: I personally prefer to syntax: 

```solidity
IUniswapExchange(tokenExchange).ethToTokenSwapInput.value(_outgoingAssetAmount)(
  _minIncomingAssetAmount,
  add(block.timestamp, 1)
);
```

Is there a reason for the using ```{ value: ...  }```?

contracts/integrations/adapters/ZeroExV2Adapter.sol
```diff
@@ -2,165 +2,165 @@
...
+                uint256 takerAssetFillAmount
+            ) = __decodeTakeOrderCallArgs(_encodedCallArgs);
+            IZeroExV2.Order memory order = __constructOrderStruct(encodedZeroExOrderArgs);
+            address makerAsset = __getAssetAddress(order.makerAssetData);
+            address takerFeeAsset = __getAssetAddress(IZeroExV2(EXCHANGE).ZRX_ASSET_DATA());
+            uint256 takerFee = __calcRelativeQuantity(
+                order.takerAssetAmount,
+                order.takerFee,
+                takerAssetFillAmount
+            ); // fee calculated relative to taker fill amount
```
fubhy, line 58: To we have a standard for how to write in-code comments?

Sometimes we end them in periods, sometimes not, sometimes we start them with a capital character, sometimes now, sometimes we write them at the end of a line, sometimes above a lines, etc.

contracts/integrations/adapters/ZeroExV3Adapter.sol
```diff
@@ -2,153 +2,158 @@
...
+            __getAssetProxy(order.takerAssetData),
+            takerAssetFillAmount
+        );
 
-        return (fillAssets, fillExpectedAmounts, fillApprovalTargets);
-    }
+        uint256 protocolFee = __calcProtocolFeeAmount();
+        if (protocolFee > 0) {
+            __increaseApproval(
+                Registry(REGISTRY).nativeAsset(),
```
fubhy, line 137: Could be `IRegistry`

contracts/integrations/utils/AdapterBase.sol
```diff
@@ -0,0 +1,178 @@
...
+            }
+        }
+    }
+
+    /// @notice Helper to determine whether an address is a valid Vault component
+    function __isVault(address _who) internal view returns (bool) {
+        // 1. Is valid Spoke of a Registered fund
+        // 2. Is the vault of the registered fund
+        try Spoke(_who).HUB() returns (address hub) {
+            return Registry(REGISTRY).fundIsRegistered(hub) && __getVault(hub) == _who;
```
fubhy, line 170: Could be `IRegistry`

contracts/integrations/utils/AdapterBase.sol
```diff
@@ -0,0 +1,178 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity 0.6.8;
+
+import "../../dependencies/token/IERC20.sol";
+import "../../dependencies/TokenUser.sol";
+import "../../fund/hub/Hub.sol";
+import "../../fund/hub/Spoke.sol";
```
fubhy: Unused imports

contracts/integrations/utils/AdapterBase.sol
```diff
@@ -0,0 +1,178 @@
...
+        for (uint256 i = 0; i < _assets.length; i++) {
+            // Ignore assets with a 0 address, 0 amount
+            if (_assets[i] == address(0) || _amounts[i] == 0) continue;
+
+            // Ignore assets that have already been added
+            bool assetAdded;
+            for (uint256 j = 0; j < i; j++) {
+                // Prev asset in array must have non-zero amount
+                if (_assets[i] == _assets[j] && _amounts[j] > 0) {
+                    assetAdded = true;
```
fubhy, line 134: Is there no `continue(2)` or so in solidity?

contracts/integrations/utils/AdapterBase.sol
```diff
@@ -0,0 +1,178 @@
...
+                    break;
+                }
+            }
+            if (!assetAdded) aggregatedAssetsCount++;
+        }
+        aggregatedAssets_ = new address[](aggregatedAssetsCount);
+        aggregatedAssetAmounts_ = new uint256[](aggregatedAssetsCount);
+        if (aggregatedAssetsCount == 0) return (aggregatedAssets_, aggregatedAssetAmounts_);
```
fubhy, line 142: How do you feel about single-line if statements in general? I usually prefer to never write body-less control statements. E.g. instead of this I'd always prefer: 

```solidity
if (aggregatedAssetsCount == 0) {
  return (aggregatedAssets_, aggregatedAssetAmounts_);
}
```

Simply because an `if` in the code immediately makes my eyes jump to the next line.

contracts/integrations/utils/AdapterBase.sol
```diff
@@ -0,0 +1,178 @@
...
+
+        uint256 aggregatedAssetIndex;
+        for (uint256 i = 0; i < _assets.length; i++) {
+            if (_assets[i] == address(0) || _amounts[i] == 0) continue;
```
fubhy, line 146: Same here for instance.

contracts/utils/MathHelpers.sol
```diff
@@ -0,0 +1,24 @@
...
+    function __calcRelativeQuantity(
+        uint256 _quantity1,
+        uint256 _quantity2,
+        uint256 _relativeQuantity1
+    )
+        internal
+        pure
+        returns (uint256 relativeQuantity2_)
+    {
+        relativeQuantity2_ = mul(_relativeQuantity1, _quantity2) / _quantity1;
```
fubhy, line 22: Why not just `return`?

contracts/integrations/adapters/UniswapAdapter.sol
SeanJCasey: The encoded call args are going to be different for each adapter, so this is validation of adapter-specific args, even if there is some overlap. There are probably some redundancies with the validation done on spendAssets and incomingAssets in Vault, but these comparisons shouldn't be expensive operations. We can further optimize later, I think.

contracts/integrations/adapters/UniswapAdapter.sol
SeanJCasey: Yes, that's now the recommended practice (you'll get a compiler warning if you use the deprecated syntax)

contracts/integrations/adapters/ZeroExV2Adapter.sol
SeanJCasey: No, that's not in the style guide. Will create a card for standardizing comments.

contracts/integrations/adapters/ZeroExV3Adapter.sol
SeanJCasey: Could be, which would make the deployment less expensive, but for now the convention is for all plugins to use full contracts for Melon core contracts. The idea is that the narrow interfaces used by our core contracts should reveal the surface area of interactions within the core system. I'm open to re-thinking the way we use interfaces, and whether plugins should use them, but for now let's stick to the convention.

contracts/integrations/utils/AdapterBase.sol
SeanJCasey: Refactored and committed

contracts/integrations/utils/AdapterBase.sol
SeanJCasey: I don't believe so (never seen it), but the next lines act on `assetAdded` being false (i.e., having gone through the entire loop and not triggering the `true` condition), so I don't think jumping to the higher for loop is relevant here? What am I missing?

contracts/integrations/utils/AdapterBase.sol
SeanJCasey: I'll create a card for updating the style guide now that there are a couple items in this PR, and we can discuss that there and then refactor.

contracts/utils/MathHelpers.sol
SeanJCasey: That would be fine too, it's the same. Just seemed like the right thing to do if we're reserving a name for the return value. This could easily have no named return value, but I think it's visually clearer to understand the function.

fubhy (approved):

PR #1005, opened 2020-06-30
refactor: imports and readme

fubhy:
This introduces a tsconfig.json purely for resolving the ~/ imports so autocomplete and code introspection works in the IDE. Also eliminates the top level `deploy` directory and instead flattens it out to the standard `migrations` directory from truffle.

Also adjusts the README according to the latest changes.


README.md
```diff
@@ -26,14 +26,8 @@ yarn install
...
-- [docker](https://www.docker.com/)
-
-After installation, go to the above `protocol` directory, have Docker running, and:
-
 ```sh
-make
+yarn compile
 ```
 
 ## Test
```
SeanJCasey, line 33: Test section still talks about Kovan...

1. remove reference to Kovan
2. add explanation that tests are intended to be run off a mainnet fork 
3. add step to add config pointing to a node (give infura as an example)... this can be part of the section in #2 above
4. get rid of "Local blockchain" header as there aren't multiple options anymore

refactor.patch
```diff
@@ -0,0 +1,7036 @@
+diff --git a/.gitignore b/.gitignore
```
SeanJCasey: What is this file for?

package.json
```diff
@@ -40,19 +40,17 @@
...
+    "jest-matcher-utils": "^26.1.0",
     "request-promise": "^4.2.2",
-    "solhint": "^2.3.0",
+    "solhint": "^3.0.0",
     "truffle": "^5.1.32",
-    "web3": "1.2.4",
-    "web3-eth-abi": "1.2.4",
-    "web3-utils": "1.2.4"
+    "ts-jest": "^26.1.1",
+    "typescript": "^3.9.5",
```
SeanJCasey, line 50: I see you've found a way to sneak typescript back in here :P 

Is this purely to do the work that babel packages were doing, or are we using TS for something else now?

SeanJCasey (changes requested):
Few minor comments/questions and a few changes to the readme to make it more accurate

package.json
fubhy: This is a replacement for babel, yes and to make the path aliasing (~/) work with IDEs autocompletion for paths, etc. Currently no further TS features (no typechecking, etc.). As mentioned before, I'd be open and willing to implement that though (and still think it would be beneficial).

refactor.patch
fubhy: Whoops. Accidentally added this one.

SeanJCasey (approved):

PR #1006, opened 2020-07-01
chore: remove unused imports

fubhy:


fubhy:
@SeanJCasey This is failing deployment on the FundFactory deployment step.

I think this is due to that one using the Hub and simply exceeding the block gas limit without optimization?

SeanJCasey (changes requested):
Let's not worry about the optimizer changes for now.

Let's create a new PR for the unused imports and updating the commit hash.

Sorry for being anal, just prefer not to have "misc" PRs.

SeanJCasey (approved):

PR #1009, opened 2020-07-01
refactor: remove specific ERC20 token contracts

fubhy:


src/dependencies/token/BurnableToken.sol
```diff
@@ -3,13 +3,10 @@ pragma solidity 0.6.8;
...
-    constructor(string memory _symbol, uint8 _decimals, string memory _name)
-        public
-        PreminedToken(_symbol, _decimals, _name)
-    {}
-
+/// @title Flexible ERC20 interface
+/// @author Melon Council DAO <security@meloncoucil.io>
+/// @dev Just a wrapper for tokens which can be burnt
+abstract contract BurnableToken is PreminedToken {
     function burn(uint _amount) public {
```
SeanJCasey: `uint256`

SeanJCasey: Also, if this contract isn't needed by the tests, let's just remove it and use a narrow interface directly in the Engine.sol contract:

```
interface IMelonToken {
    function burn(uint256) external;
}
```

Does Truffle compile interfaces? If so, we could also create an `contract/interfaces/` dir, so as not to have conflicts if we need a Melon token interface elsewhere.

SeanJCasey (changes requested):

SeanJCasey (approved):

PR #1011, opened 2020-07-01
chore: change package.json author

fubhy:


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@melonproject/protocol",
   "version": "1.0.12",
   "description": "Technology Regulated and Operated Investment Funds",
-  "author": "Melonport AG <team@melonport.com>",
+  "author": "Melon Council DAO <team@meloncoucil.io>",
```
SeanJCasey: "council" is misspelled; `team@meloncouncil.io`

Also, is this the correct address? We're using `@author Melon Council DAO <security@meloncoucil.io>` in the Natspec of every contract (which I now realize is misspelled also).

SeanJCasey (changes requested):

package.json
fubhy: Yeah for security that's correct. The team@ address also exists tho.

fubhy: I copied the meloncoucil error from there btw. :-D

SeanJCasey (approved):

PR #1013, opened 2020-07-01
feat: test utils and abi code generator

fubhy:
Done:
- [x] Generate typescript classes for ethers.js
- [x] Add ethers.js providers and signers using ganache state
- [x] Add ganache provider with access to ganache state trie for send / call assertions
- [x] Add utils for test fixtures (fund setup, etc.)

Next up (follow up):
- [x] Add assert utils for events
- [x] Add assert utils for send / call code paths (unit tests)
- [x] Add better matchers for big number, etc.

... More to be added


fubhy:
@SeanJCasey Let's merge this <3

SeanJCasey (approved):

PR #1014, opened 2020-07-02
test: add a custom test env with a globally shared web3 provider

fubhy:
- [x] Upgraded to Node v12
- [x] Added a custom jest test environment (based on the default node environment) that spins up the ganache fork for every test file and allocates a shared global for it removing the need to pass it around everywhere
- [x] Shared ganache configuration
- [x] Fixed a race condition in the amgu test


fubhy:
This is the foundation for further work on https://github.com/melonproject/protocol/pull/1013

SeanJCasey (approved):

PR #1016, opened 2020-07-08
fix: kyber pricefeed missing functions

SeanJCasey:
The primary purpose of this PR is to re-implement a function that is defined in `IPriceSource.sol` and is used in `Participation.sol`, but is missing from `KyberPriceFeed.sol`, which caused the price feed to be unusable.

Additionally, it:
- makes `KyberPriceFeed.sol` inherit from `IPriceSource.sol`, and re-implements all other missing functions (even though these other functions are not used in the protocol codebase)
- reverts name of an event to its old name, to be consistent with the production version of the Melon subgraph
- replaces `_failIfInvalid` flag on `update()` with check that allows a 0 price to pass if the sanity rate (i.e., its expected off-chain rate) is exactly 0 (e.g., SAI does not have a Kyber reserve price, but it is in the list of registered assets)
- replaces the use of arbitrary `reserveMin` values in passing a `srcQty` to `KyberNetworkProxy.getExpectedRate()` with dynamic calculations of the src asset relative to a constant amount of weth (e.g., 0.01 WETH worth of DAI, 0.01 WETH worth of WBTC, etc).

### DIFF
--- a/src/prices/KyberPriceFeed.sol
+++ b/src/prices/KyberPriceFeed.sol
@@ -4,11 +4,12 @@ import "../dependencies/DSMath.sol";
 import "../dependencies/token/IERC20.sol";
 import "../exchanges/interfaces/IKyberNetworkProxy.sol";
 import "../registry/IRegistry.sol";
+import "./IPriceSource.sol";
 
 /// @title Price Feed Template
 /// @author Melon Council DAO <security@meloncoucil.io>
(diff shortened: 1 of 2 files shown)


fubhy (approved):

fubhy:
Closing this as we agreed to not pursue this further before the upcoming release. Will leave the branch in place in case we decide otherwise.

PR #1029, opened 2020-08-04
refactor: immutable vars

luongnt95:
@SeanJCasey Maybe consider using "WETH_TOKEN" or "NATIVE_ASSET_TOKEN" for naming variables and external functions

### DIFF
--- a/contracts/registry/Registry.sol
+++ b/contracts/registry/Registry.sol
@@ -51,6 +51,10 @@ contract Registry is MelonCouncilOwnable {
 
     event ValueInterpreterChanged(address valueInterpreter);
 
+    // Constants
+    address immutable public MLN_TOKEN;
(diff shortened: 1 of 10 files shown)


contracts/registry/Registry.sol
```diff
@@ -72,6 +68,9 @@ contract Registry is MelonCouncilOwnable {
     mapping (address => bool) public fundFactoryIsRegistered;
 
     // Funds
+    address immutable public MLN_TOKEN;
```
SeanJCasey: Let's move both of these above the storage vars, under a `// Constants` comment (L54, before `// Assets`).

SeanJCasey (changes requested):
Nice job @luongnt95 ! Just one small formatting change, then we can get this merged!

SeanJCasey (approved):

PR #1030, opened 2020-08-04
chore: automatically publish dev packages

fubhy:


fubhy:
Automatically publishes packages like so: `2.0.0-dev.$commithash`: https://www.npmjs.com/package/@melonproject/protocol?activeTab=versions

SeanJCasey (approved):

PR #1031, opened 2020-08-05
chore: fix solhint warnings and disallow future warnings

fubhy:
This fixes all the warnings we have left currently.

Note the `solhint-disable-line` statements for non-issues e.g. the reentrancy one in SharesRequestor which is incorrectly flagging that line because of `msg.sender.transfer` in a different code branch (if/else statement above so not relevant)

I've also set `max-warnings` to 0 so that it will fail from now on if new warnings are introduced.

### DIFF
--- a/contracts/fund/policies/utils/AddressListPolicyMixin.sol
+++ b/contracts/fund/policies/utils/AddressListPolicyMixin.sol
@@ -13,7 +13,7 @@ abstract contract AddressListPolicyMixin {
 
     event AddressesRemoved(address policyManager, address[] items);
 
-    mapping (address => EnumerableSet.AddressSet) policyManagerToList;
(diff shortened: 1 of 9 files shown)


contracts/fund/policies/utils/AddressListPolicyMixin.sol
```diff
@@ -13,7 +13,7 @@ abstract contract AddressListPolicyMixin {
 
     event AddressesRemoved(address policyManager, address[] items);
 
-    mapping (address => EnumerableSet.AddressSet) policyManagerToList;
+    mapping (address => EnumerableSet.AddressSet) private policyManagerToList;
```
SeanJCasey, line 16: This is fine, but it could arguably be `internal` since it's an abstract contract that is only inherited (i.e., a child may have some use for it).

SeanJCasey (approved):

PR #1037, opened 2020-08-12
feat: adapter policies

luongnt95:


tests/tests/AdapterBlacklist.test.ts
```diff
@@ -0,0 +1,127 @@
...
+      });
+
+      const encodedArgs = await engineTakeOrderArgs(1, mlnAmount);
+      tx = fund.vault.callOnIntegration(
+        engineAdapter,
+        takeOrderSignature,
+        encodedArgs,
+      );
+
+      // this revert message is reached only when the pre-validate policies pass
```
fubhy: in that case, instead of building the test around this revert statement, we should probably try to actually make it go through.

that would mean, in this case, to actually send some eth into the engine.

tests/tests/AdapterWhitelist.test.ts
```diff
@@ -0,0 +1,127 @@
...
+      });
+
+      const encodedArgs = await engineTakeOrderArgs(1, mlnAmount);
+      tx = fund.vault.callOnIntegration(
+        engineAdapter,
+        takeOrderSignature,
+        encodedArgs,
+      );
+
+      // this revert message is reached only when the pre-validate policies pass
```
fubhy: same here.

fubhy (changes requested):

tests/utils/fund/engine.ts
```diff
@@ -0,0 +1,55 @@
+import { BigNumberish, Signer } from 'ethers';
```
fubhy: This file should not be in the `fund` directory. The `fund` directory is for utilities around fund creation, etc. This one could just be in `utils/engine.ts` and exported from `utils/index.ts` too.

fubhy (changes requested):

tests/contracts.ts
```diff
@@ -32,6 +32,8 @@ export { ZeroExV3Adapter } from '../codegen/ZeroExV3Adapter';
 
 // Policies
 export { UserWhitelist } from '../codegen/UserWhitelist';
+export { AdapterBlacklist } from './codegen/AdapterBlacklist';
```
fubhy: This is in ../codegen now

fubhy (changes requested):

package.json
```diff
@@ -36,34 +37,23 @@
     "@openzeppelin/contracts": "^3.1.0"
   },
   "devDependencies": {
-    "@crestproject/crestproject": "^0.0.23",
```
fubhy, line 39: I did a bunch of updates to dependencies here while debugging the test failure. Fixed the failure in crestproject but keeping these updates is not bad either way.

tests/deployment.ts
```diff
@@ -420,9 +432,11 @@ export async function defaultTestConfig(
   ]);
 
   // Send weth to each exchange.
-  await exchanges.map((exchange) => {
-    return weth.transfer(exchange, utils.parseEther('100'));
-  });
+  await Promise.all(
```
fubhy, line 435: This is a fix for a bug that I added in a previous commit accidentally. This did not manifest itself in a test failure because it's probably a rare race condition but I stumbled over this and thought fixing it here doesn't hurt.

contracts/fund/policies/risk-management/AdapterBlacklist.sol
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity 0.6.8;
+
+import "../utils/CallOnIntegrationPreValidatePolicyBase.sol";
+import "../utils/AddressListPolicyMixin.sol";
+
+/// @title AdapterBlacklist Contract
+/// @author Melon Council DAO <security@meloncoucil.io>
+/// @notice A blacklist of adapters to add to a fund's vault
+/// @dev Adapters can be added but not removed from blacklist
```
SeanJCasey: This is no longer true. The blacklist config can't be changed.

Can you please remove this comment from AssetBlacklist also?

contracts/fund/policies/risk-management/AdapterWhitelist.sol
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity 0.6.8;
+
+import "../utils/CallOnIntegrationPreValidatePolicyBase.sol";
+import "../utils/AddressListPolicyMixin.sol";
+
+/// @title AdapterWhitelist Contract
+/// @author Melon Council DAO <security@meloncoucil.io>
+/// @notice A whitelist of adapters to add to a fund's vault
+/// @dev Adapters can be removed but not added from whitelist
```
SeanJCasey: This is no longer true. The whitelist config can't be changed.

Can you please remove this comment from AssetWhitelist also?

SeanJCasey (changes requested):
Looks good!

Just a couple small changes to comments.

PR #1055, opened 2023-10-07
fix: Makefile error :79 [unexpected token] fixed

milaabl:
This line fixes the `make build` error for me in a Command prompt:

**OS:** _Windows 11_

```bash
Generating interfaces ...
)nvalid line format in tests/interfaces/interfaces.txt (
make: *** [Makefile:79: tests/interfaces/internal/] Error 1
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ $(ARTIFACTS_DIR)/: Makefile $(shell find $(CONTRACTS_DIR) -type f -name "*.sol")
 > $(FORGE) build --sizes --extra-output-files abi
 > touch $@
 
-$(INTERFACES_DIR)/: Makefile $(INTERFACES_FILE) $(ARTIFACTS_DIR)/
+$(INTERFACES_DIR): Makefile $(INTERFACES_FILE) $(ARTIFACTS_DIR)/
 > mkdir -p $(@D)
 >
 > # Remove all existing interfaces and abis.


fubhy:
Hi. Thanks for the PR. This fix should not be necessary (and is also not corect because the `/` is required to explicitly identify the path as a directory). What version of `make` are you using if I may ask?

milaabl:
@fubhy Ah, got it, thank you! I'm using `Make 4.4.1`

fubhy:
I think I can fix that by instead using a sentinel file. Let me try ...

fubhy:
This should fix it: https://github.com/enzymefinance/protocol/commit/d91720e4140bb623327df150be9096ebff59ae7a

milaabl:
Hmm, Sentinel didn't fix that for me for some reason

fubhy:
Hmm but it doesn't have the `/` at the end anymore.

fubhy:
The error also sounds like it's an issue with interfaces.txt and not the Makefile

fubhy:
https://github.com/enzymefinance/protocol/blob/d91720e4140bb623327df150be9096ebff59ae7a/Makefile#L94-L100

milaabl:
@fubhy yeah, that's indeed strange, might be something with file encodings on my end maybe. OK, thank you anyways!! I'll just keep the update that makes it works for me on my local end then.

fubhy:
> @fubhy yeah, that's indeed strange, might be something with file encodings on my end maybe. OK, thank you anyways!! I'll just keep the update that makes it works for me on my local end then.

The problem is that your change merely disables a part of the make file by breaking the reference to the interfaces build step. It is not fixing it but merely short circuiting it heh

PR #1062, opened 2025-08-08
Update VaultLib.sol

syed-ghufran-hassan:
Currently there's no event emitted when asset approvals are set via __approveAssetSpender so I have added the event


SeanJCasey:
we're not accepting patches; code is frozen, thanks

